PR #27, opened 2012-06-09
Added serialization for udtf and lvj OI's in LateralViewJoinOperator + lateral view tests. fix for #25. 

harveyfeng:
In LateralViewJoinOperator.scala, used Kryo to serialize udtf and lvf OI's and converted the byte arrays to strings so the @BeanProperty keyword/XML serialization is less inefficient. 
Also removed some unused imports.

Passed these tests on cluster and local computer that use lateral view:
testCliDriver_lateral_view
testCliDriver_lateral_view_ppd
testCliDriver_load_dyn_part15 (originally failed in issue #25)
testCliDriver_udtf_json_tuple
testCliDriver_udtf_parse_url_tuple

Added to tests_pass.txt:
testCliDriver_lateral_view
testCliDriver_lateral_view_ppd 
testCliDriver_udtf_json_tuple
testCliDriver_udtf_parse_url_tuple

### DIFF
--- a/src/main/scala/shark/exec/LateralViewJoinOperator.scala
+++ b/src/main/scala/shark/exec/LateralViewJoinOperator.scala
@@ -2,6 +2,7 @@ package shark.exec
 
 import java.util.ArrayList
 
+import org.apache.commons.codec.binary.Base64
 import org.apache.hadoop.hive.ql.exec.{ExprNodeEvaluator, ExprNodeEvaluatorFactory, LateralViewJoinOperator => HiveLateralViewJoinOperator}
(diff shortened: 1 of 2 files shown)


rxin:
What'd happen if the byte array contains value 0?

It might be safer to use base64 encoding here.

harveyfeng:
0 would encode a n ascii null terminator (
http://www.tachyonsoft.com/iso88591.htm)

It's a one to one mapping for all byte values, so it should work. I'll
patch this with base64 to be more consistent with Hive encodings...

On Sun, Jun 10, 2012 at 1:47 PM, Reynold Xin <
reply@reply.github.com

PR #34, opened 2012-07-30
JUnit

dennybritz:
Runs The Hive tests using Ant and JUnit in order to produce an XML test report.


rxin:
I merged this offline into the dev branch. Thanks!

PR #37, opened 2012-08-16
Check for HIVE_DEV_HOME specified w/ tests.

pwendell:
The test runner currently NPE's if HIVE_DEV_HOME is not set. This adds
a check for that case in the sbt run script to give a clearer error
message.


rxin:
Thanks!

PR #46, opened 2012-09-11
Cache tables using tblproperties; determine initial columnar cache ArrayList capacity during runtime and provide a config for it.  

harveyfeng:
Also some formatting

### DIFF
--- a/src/main/scala/shark/ColumnarSerDe.scala
+++ b/src/main/scala/shark/ColumnarSerDe.scala
@@ -4,17 +4,25 @@ import java.io.{DataInput, DataOutput}
 import java.util.Properties
 
 import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.hdfs.DFSConfigKeys
 import org.apache.hadoop.hive.serde2.{ByteStream, SerDe, SerDeException}
 import org.apache.hadoop.hive.serde2.`lazy`.{LazyFactory, LazySimpleSerDe}
 import org.apache.hadoop.hive.serde2.`lazy`.LazySimpleSerDe.SerDeParameters
+import org.apache.hadoop.hive.serde2.objectinspector.ListObjectInspector
+import org.apache.hadoop.hive.serde2.objectinspector.MapObjectInspector
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector
+import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector.Category
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorUtils
+import org.apache.hadoop.hive.serde2.objectinspector.primitive.PrimitiveObjectInspectorFactory
+import org.apache.hadoop.hive.serde2.objectinspector.PrimitiveObjectInspector
+import org.apache.hadoop.hive.serde2.objectinspector.PrimitiveObjectInspector.PrimitiveCategory
 import org.apache.hadoop.hive.serde2.objectinspector.StructField
 import org.apache.hadoop.hive.serde2.objectinspector.StructObjectInspector
-import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector.Category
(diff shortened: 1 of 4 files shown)


src/main/scala/shark/ColumnarSerDe.scala
```diff
@@ -82,14 +91,26 @@ class ColumnarSerDe extends SerDe {
   }
 }
 
-
-class ColumnarWritable(oi: StructObjectInspector) extends Writable {
+class ColumnarWritable(oi: StructObjectInspector, conf: Configuration)
```
rxin: I'd pass the estimated block size in instead of the entire conf and just determine the size in ColumnSerDe.initialize.

And add a comment on "we are approximating partition size based on hdfs block size, etc"

src/main/scala/shark/ColumnarSerDe.scala
```diff
@@ -110,6 +131,76 @@ class ColumnarWritable(oi: StructObjectInspector) extends Writable {
...
+
+  // Sizes of primitive types
+  private val BOOLEAN_SIZE = 1
+  private val BYTE_SIZE = 1
+  private val SHORT_SIZE = 2
+  private val INT_SIZE = 4
+  private val LONG_SIZE = 8
+  private val FLOAT_SIZE = 4
+  private val DOUBLE_SIZE = 8
+  private val STRING_SIZE = 16
```
rxin: Add a comment on we are using heuristics (16 bytes) to determine string size ...

src/main/scala/shark/ColumnarSerDe.scala
```diff
@@ -110,6 +131,76 @@ class ColumnarWritable(oi: StructObjectInspector) extends Writable {
...
+
+  // Size of an object reference.
+  private val pointerSize = if (System.getProperty("os.arch").contains("64")) 8 else 4
+
+  // Estimates for complex types. Includes overhead and pointer sizes for Java HashMap and
+  // ArrayList objects, based on Spark's SizeEstimator.estimate().
+  private val avgListLength = 5
+  private val listOverhead = 88 + pointerSize * avgListLength
+  private val avgMapSize = 5
+  private val mapOverhead = 152 + pointerSize * avgMapSize
```
rxin: The serialized form probably doesn't contain these overheads right? Map/lists are serialized using the serde to cache.

src/main/scala/shark/ColumnarSerDe.scala
```diff
@@ -82,14 +91,26 @@ class ColumnarSerDe extends SerDe {
...
 
+  // Determine the initial capacity for ArrayList columns.
+  var initialColumnSize = SharkConfVars.getIntVar(conf, SharkConfVars.COLUMN_INITIALSIZE)
+  if (initialColumnSize == - 1) {
+    // Try both "dfs.block.size" and "dfs.blocksize" (from DFS_BLOCK_SIZE_KEY),
+    // which is the new name since Hadoop 0.21.0.
+    val partitionSize = conf.getLong(DFSConfigKeys.DFS_BLOCK_SIZE_KEY,
+      conf.getLong("dfs.block.size", DFSConfigKeys.DFS_BLOCK_SIZE_DEFAULT))
+    val fieldSize = ColumnarWritable.getFieldSize(oi)
+    initialColumnSize = (partitionSize / (fieldSize *
```
rxin: Add a high level comment on how we estimate initial size.

rxin:
Nicely done, Harvey.

I added a few comments - mostly on adding a bit of documentation.

Also, for the _cached part, can we introduce one config variable that enables the _cached behavior too? This can maintain backwards compatibility, also _cached (although hacky) is actually easier to write than table properties.

So if shark.cache in table properties is set, it is always going to cache it. But if conf variable "shark.underscore_cached" is set, it's also going to cache tables that end with _cached.

Thanks!

shark.underscore_cached is a bad name. If you can come up with a better name, I am all for it.

rxin:
Merged. Thanks!

PR #30, opened 2012-07-06
Migrating to Hive 0.9.0

dennybritz:
- A couple of minor changes in package/class naming
- SerDe Stats does not collect any actual stats yet. This is a TODO.
- ExplainWork formatted (JSON) output disabled by constructor
- Fixes #24

PR #54, opened 2012-09-28
Changes for Hive 0.9

harveyfeng:
Additions to pull request/issue #30


harveyfeng:
Tested the CLI and server with same commands from https://reviews.apache.org/r/407/

```
- set; -- get all parameters (called multiple times to check for IO reset)
- set hive.stats.autogather;  -- check default parameter value
- set hive.stats.autogather=false;  -- change parameter value
- set hive.stats.autogather;  -- check parameter value got changed
- select * from src;  -- Shark query + Spark job
- select count(*) from src; -- Shark query + Spark job
- select k from src; -- negative test case where SemanticAnalyzer throws an exception. Tests if error messages are passed to client.
- show partitions srcpart;  -- query, but no job
- explain select count(*) from srcpart where ds is not null; -- explain query output
```

rxin:
This has been merged into hive0.9 branch.

PR #61, opened 2012-10-04
Register Shark's conf variables with SessionState Configuration

harveyfeng:
Previously, couldn't use "set <conf var>" to see default values of Shark's conf vars, which were also missing from "set;":

shark> set shark.explain.mode;
shark.explain.mode is undefined


rxin:
Thanks - terrific.

PR #65, opened 2012-10-10
Add stop() method to SparkEnv

dennybritz:


rxin:
Can you add a shutdown hook to the server too?

dennybritz:
Done.

PR #76, opened 2012-11-06
Add table property and conf var for RDD storage level.

harveyfeng:
Example:
CREATE TABLE src_cached  
TBLPROPERTIES ("shark.cache"="true", "shark.cache.storageLevel"="MEMORY_AND_DISK")
AS SELECT \* FROM src;


rxin:
Thanks!

PR #77, opened 2012-11-15
Initial cut for Shark-64

harsha2010:
https://spark-project.atlassian.net/browse/SHARK-64
This checkin fixes Shark-64.
When cached tables are created, meta data is added to the hive meta store for each table, that includes the query that resulted in the cached table creation.
Upon server restarts, we run through the queries to re populate the cached RDDs before resuming work.
Also added a few JDBC tests for Shark Server

### DIFF
--- a/src/main/scala/shark/SharkServer.scala
+++ b/src/main/scala/shark/SharkServer.scala
@@ -5,10 +5,8 @@ import java.io.IOException
 import java.io.PrintStream
 import java.io.UnsupportedEncodingException
 import java.util.ArrayList
+import java.util.concurrent.CountDownLatch
 import java.util.{List => JavaList}
-
-import scala.concurrent.ops.spawn
-
 import org.apache.commons.logging.LogFactory
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.hive.common.LogUtils
@@ -28,9 +26,9 @@ import org.apache.thrift.server.TThreadPoolServer
 import org.apache.thrift.transport.TServerSocket
 import org.apache.thrift.transport.TTransport
 import org.apache.thrift.transport.TTransportFactory
-
+import scala.annotation.tailrec
+import scala.concurrent.ops.spawn
 import spark.SparkEnv
-
 
 object SharkServer {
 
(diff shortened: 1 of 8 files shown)


src/main/scala/shark/SharkServer.scala
```diff
@@ -43,24 +58,29 @@ object SharkServer {
...
+      while(!server.isServing()) {}
+      val sshandler = new SharkServerHandler
+      SharkCTAS.loadAsRdds(sshandler.execute(_))
+    }
     LOG.info("Starting shark server on port " + port)
     server.serve()
   }
-
+  
+  def stop = if(server !=null) server.stop
```
rxin: nitpick: a space after if.

src/main/scala/shark/SharkServer.scala
```diff
@@ -43,24 +58,29 @@ object SharkServer {
...
-          SharkEnv.stop()
+          SharkServer.stop
         }
       }
     )
 
+    spawn {
+      while(!server.isServing()) {}
+      val sshandler = new SharkServerHandler
+      SharkCTAS.loadAsRdds(sshandler.execute(_))
```
rxin: What's the behavior if users are querying these cached tables when the server is first started? Should we block the server when it is creating cached tables?

harsha2010:
Thanks Reynold, will fix the space and recommit.
Yes, you are right, we will need to block the server when it is creating cached tables. I will add that in and check in

rxin:
BTW - it might be better to have a flag in the server/cli start script to
disable the loading, in case the loading itself can deterministically go
wrong.

## 

Reynold Xin
AMPLab, UC Berkeley
http://www.cs.berkeley.edu/~rxin/

harsha2010:
good point, will include it. I also need to harden the code to deal with partial failures : if reloading one of several cached tables fail, we should log it and move on, i would think..but the patch exits eagerly.

src/main/scala/shark/SharkServer.scala
```diff
@@ -27,46 +38,83 @@ object SharkServer {
...
       }
     )
-
+    execLoadRdds(loadRdds, latch)
     LOG.info("Starting shark server on port " + port)
     server.serve()
   }
-
+  
+  def stop = if (server != null) {
```
rxin: Also need to call SharkEnv.sc.stop() here.

src/main/scala/shark/SharkServer.scala
```diff
@@ -27,46 +38,83 @@ object SharkServer {
...
+      try {
+        val sshandler = new SharkServerHandler
+        SharkCTAS.loadAsRdds(sshandler.execute(_))
+        LOG.info("Executed load "+ SharkCTAS.getMeta)
+      } catch {
+        case (e: Exception) => LOG.warn("Unable to load RDDs upon startup", e)
+      } finally {
+        latch.countDown
+      }
+  }
```
rxin: The brackets here are correct, but look a little bit confusing. Can we add more parenthesis to make it more clear? 

E.g.

```
if (!loadFlag) { 
  latch.countDown
} else spawn {
  ...
}
```

src/main/scala/shark/memstore/CacheManager.scala
```diff
@@ -23,4 +23,10 @@ class CacheManager {
     keyToRdd.keys.map(_.key).collect { case k: String => k } toSeq
   }
 
+  def drop(): Seq[String] = {
```
rxin, line 26: Is this only used for testing? If yes, can we add a comment saying that?

rxin:
Closing this one because I am merging it manually.

PR #78, opened 2012-11-26
Support 'insert' updates for cached tables (Shark-93)

harveyfeng:
For "insert into" commands, a UnionRDD combines the "select" query output with the table RDD being updated. For "insert overwrite", the output RDD replaces the old one in CacheManager.

Included tests:
- "insert into" on a single cached table
- "insert overwrite" on a single cached table
- "insert" on a cached table in a command w/ multiple inserts throws an error


src/main/scala/shark/parse/SharkSemanticAnalyzer.scala
```diff
@@ -141,48 +143,62 @@ class SharkSemanticAnalyzer(conf: HiveConf) extends SemanticAnalyzer(conf) with
...
+          }
+        }
+      } else if (hiveSinkOps.size == 1) {
+        // For a single output, we have the option of choosing the output
+        // destination (e.g. CTAS with table property "shark.cache" = "true").
+        Seq {
+          if (qb.isCTAS && qb.getTableDesc != null && shouldCache) {
+            val storageLevel = qb.getTableDesc.getTblProps.getOrElse("shark.cache.storageLevel",
+              SharkConfVars.getVar(conf, SharkConfVars.STORAGE_LEVEL)).toUpperCase match {
+                case "NONE" => StorageLevel.NONE
```
rxin: Can we move this match into a static function in CacheManager?

harveyfeng: Will do.

harsha2010:
Hi Harvey

It appears this patch interferes with stats collection.
https://github.com/harveyfeng/shark/blob/cache-insert/src/main/scala/shark/execution/TerminalOperator.scala#L219

In line # 219, the stats are prematurely collected before the RDD serialization code is executed. The fix is to first execute rdd.foreach(_ => Unit) and then collect stats and store.

harveyfeng:
Sry about that. It's been updated.

harsha2010:
Thanks Harvey

rxin:
Closing this pull request because I am manually merging this into the Tachyon branch, and it will find its way into master soon.

rxin:
The change has been merged into the tachyon branch.

PR #80, opened 2012-11-28
Pass missing env vars needed for distributed mode

harveyfeng:
SPARK_HOME, SCALA_HOME and MESOS_NATIVE_LIBRARY weren't being passed to slaves


rxin:
Thanks!

PR #81, opened 2012-12-03
Shark-64 cleanup, add documentation and fix test

harsha2010:
refactor classname to CachedTableRecovery, add some documentation on methods, fix test to correctly throw up exception when it fails.


rxin:
Thanks! Looks good.

I am going to change the tests a little bit to provide better isolation between different test suites (i.e. using different metastore and warehouse dirs).

rxin:
BTW - the related commit is: https://github.com/amplab/shark/commit/d45e58c5929402306f973d6e40185c972434c78b

PR #82, opened 2012-12-06
Add Apache license to all source files

pwendell:
These commits add the Apache 2.0 License (LICENSE) file and adds headers to all source files.


rxin:
Thanks!

PR #79, opened 2012-11-27
Maven POM file for Shark, Debian packaging, and "run" script improvement

mbautin:
- Adding a POM file for building with Maven, both with Hadoop 1 and Hadoop 2
- Adding Debian packaging
- Updating the run script to be able to run with a fat jar (no Scala
  dependency).

PR #83, opened 2012-12-18
Maven POM file and Debian packaging (on master)

mbautin:
Submitting https://github.com/amplab/shark/pull/79 on top of the master branch. Hive version is now specified as 0.9.0, so it should build out of the box. However, to use Shark-patched Hive one would have to specify -Dhive.version explicitly.


massie:
Thanks for submitting this, @mbautin.  I'll take a look at it soon.  Looking forward to moving Shark over to Maven.

rxin:
Hi Mikhail,

I start to look at the build file changes and your Hive patches. If we are publishing the amplab Hive to Maven, perhaps we can get rid of the HIVE_HOME environmental variable, and only keep HIVE_CONF_DIR?

mbautin:
Hi Reynold, I think HIVE_HOME is still needed to locate Hive jars and add them to the classpath (correct me if I'm wrong).

mbautin:
@rxin @massie any chance this could be integrated soon?

harsha2010:
Hey Mikhael,

I tried running a maven build with your pom, and am running into two issues : 
1. hive-exec-test-0.9.jar seems to be needed (this can be locally built by running an ant target in hive directory, but probably needs to be documented unless i am missing something)
2. hadoop-core-0.20.1 seems to be required and cannot be found in the repos.
Failed to execute goal on project shark: Could not resolve dependencies for project edu.berkeley.cs.amplab:shark:jar:0.3.0-SNAPSHOT: Failure to find org.apache.hadoop:hadoop-core:jar:0.20.1 in http://repo.typesafe.com/typesafe/releases/ was cached in the local repository, resolution will not be reattempted until the update interval of TypesafeRepository has elapsed or updates are forced

mbautin:
Hi Ram,

hive-exec-test.jar should only be required for testing, and it has to be built locally anyway, as Hive's auto-generated test code contains hard-coded paths. However, we do need to publish our patched Hive jars somewhere. I opened an issue with Sonatype to push custom builds of Hive from the codebase at https://github.com/amplab/hive to the Maven central repository: https://issues.sonatype.org/browse/OSSRH-5250

I will look at the hadoop-core issue shortly.

Thanks,
Mikhail

mbautin:
Ram: I modified the Hive version to the one built from the code at https://github.com/amplab/hive deployed to a temporary repository location at http://people.apache.org/~mbautin/maven/. Does AMP Lab have a public Maven repository server? Ideally these builds could be pushed there. Could you please try building Shark again and let me know what happens?

Also, I haven't observed a problem with hadoop-core-0.20.1. Could you please let me know the exact command line you are using to build Shark?

mag-:
Any chances for merging this before next release? maven build with debian package is really handy

jey:
Hi Mikhail,

Thanks for this PR, I'm currently reviewing it for inclusion. However I have a couple of questions, and please bear with me if my ignorance of Maven is at fault here.
1. Could you please rebase your branch to the latest shark/master and confirm that it still builds and passes tests?
2. The following additional patch seems necessary: https://gist.github.com/jey/5138295
3. Even with the above patch I'm getting dependency failures when Maven tries to satisfy the dependency on Spark. Since we don't yet have a maven repository with Spark binaries, what's the correct way to point Maven at my local Spark installation?

Thanks,
-Jey

mbautin:
Hi Jey,

Thank you for looking at this pull request. It would certainly be great to get it committed. I will rebase the pull request, apply your patch, and run unit tests.

To be able to use locally-built Spark 0.7.0, you can do the following:

```
cd spark
git checkout v0.7.0
find . -name "pom.xml" -exec sed -i "" 's/0.7.0-SNAPSHOT/0.7.0/g' {} \;
mvn -Phadoop2 install
```

The `find . -name "pom.xml" -exec sed -i "" 's/0.7.0-SNAPSHOT/0.7.0/g' {} \;` is needed because in the git revision of Spark labeled as `v0.7.0` submodules still refer to version `0.7.0-SNAPSHOT` of the parent POM. (The standard way to release projects in Maven is `mvn clean release:prepare release:perform`, which takes care of these versions. Not sure if it was used when releasing Spark 0.7.0.)

After this you should be able to build Shark. I will update the pull request with the updated version shortly, start unit tests, and post another update when I get the results.

Thanks,
Mikhail

jey:
Thanks, much appreciated. I'm also working on getting the Maven repository updated with Spark binaries that explicitly specify their hadoop version, so that dependencies such as "org.spark-project:spark-core:jar:hadoop2:0.7.0" will be automatically resolved when building Shark via Maven.

mbautin:
@jey: I have updated the pull request, built Spark 0.7.0 as described above, and verified that Shark compilation also works once Spark jars have been installed into the local Maven repository. Please let me know if it works in your environment as well. I am running Shark unit tests, and will post and update when I get the results.

mbautin:
@jey: here is the unit test output: https://gist.github.com/mbautin/aec2c99df9f9837b762d
26 out of 707 TestCliDriver tests failed (consistent with my previous runs), and one of scalatest tests failed as well.

Failed TestCliDriver tests:

```
Tests run: 707, Failures: 26, Errors: 0, Skipped: 0, Time elapsed: 1,271.972 sec <<< FAILURE!
```

Failed scalatest tests:

```
shark.CachedSuite.Cached table with simple types

assertion failed: Didn't find "Time taken" in the output: create table shark_test1(key int, val string);
Stacktrace

      java.lang.AssertionError: assertion failed: Didn't find "Time taken" in the output:
create table shark_test1(key int, val string);
      at shark.CliTestToolkit$class.waitForQuery(CliTestToolkit.scala:64)
      at shark.CachedSuite.waitForQuery(CachedSuite.scala:24)
      at shark.CliTestToolkit$class.executeQuery(CliTestToolkit.scala:56)
      at shark.CachedSuite.executeQuery(CachedSuite.scala:24)
      at shark.CachedSuite$$anonfun$1.apply$mcV$sp(CachedSuite.scala:51)
      at shark.CachedSuite$$anonfun$1.apply(CachedSuite.scala:49)
      at shark.CachedSuite$$anonfun$1.apply(CachedSuite.scala:49)
      at org.scalatest.FunSuite$$anon$1.apply(FunSuite.scala:1265)
```

We are planning to look into these test failures. I am not sure if these tests succeed when running with sbt, but this probably should not block us from merging the Maven build file, because the majority of tests do succeed.

PR #84, opened 2012-12-18
Adding TaskContext to compute methods in Shark RDDs (patch by Thomas Dudziak)

mbautin:
This is necessary for Shark to build against the Spark master branch.


rxin:
Thanks, Mikhail / Thomas!

PR #88, opened 2013-01-22
Destroy driver properly in sql2rdd to prevent leaking resources

mbautin:


rxin:
Thanks. Quick question - is the semantics in this try/final an exception will be thrown upstream to the caller of sql2rdd in the case of failures?

mbautin:
Yes. I think if an exception is thrown both in `try` and in `finally`, the one thrown from `finally` will be suppressed and it's up to the user to handle it. However, this fixes a much more serious resource leak that happens after less than a hundred `sql2rdd` queries.

rxin:
Thanks.

mbautin:
BTW (off-topic), are you guys planning to take a look at https://github.com/amplab/shark/pull/83? It's been a couple of months :)

rxin:
Yes! Matt has been busy with recruiting, but I just recruited @harsha2010 to review it.

PR #92, opened 2013-05-10
Fix HBase filter pushdown

mbautin:


mbautin:
@rxin good point, submitted a pull request.

rxin:
Thanks - let me know if you have other things like this too :)

src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -248,6 +249,11 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
     if (tableDesc != null) {
       Utilities.copyTableJobPropertiesToConf(tableDesc, conf)
     }
+    new HiveInputFormat() {
+      def doPushFilters() {
+        pushFilters(conf, hiveOp)
+      }
+    }.doPushFilters()
```
rxin, line 256: What is the difference between this and directly calling 

pushFilters(conf, hiveOp)

?

rxin:
Ah never mind my question. You need an object to get access to the push filters method.

PR #94, opened 2013-05-14
code gen for expression eval

chenghao-intel:
And I also attached a document at https://spark-project.atlassian.net/browse/SHARK-73


rxin:
Thanks! This is very exciting. I will take a look at this asap.

Do you have any numbers on the overhead of invoking the Java compiler, e.g. how long does it take to compile the generated code down to byte code?

chenghao-intel:
I just did an micro-benchmark with my desktop, it took about 20 seconds to compile a complex generated source and then load the class for 1000 times.

It may impacts the streaming process, but I has put the improvements for the future work.

Thanks,

src/main/scala/shark/execution/cg/CGEvaluatorFactory.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.hadoop.hive.ql.exec.ExprNodeFieldEvaluator
+import org.apache.hadoop.hive.ql.exec.ExprNodeNullEvaluator
+import org.apache.hadoop.hive.ql.exec.ExprNodeEvaluatorFactory
+import org.apache.hadoop.hive.conf.HiveConf
+
+import shark.SharkConfVars
+import shark.LogHelper
+
+object CGEvaluatorFactory extends LogHelper {
+  def get(desc : ExprNodeDesc, hconf:HiveConf) : ExprNodeEvaluator = {
```
rxin: Maybe just take a boolean value instead of hconf? This way, we don't need to pass the hconf to every slave node.

src/main/scala/shark/execution/cg/CodeGenManager.scala
```diff
@@ -0,0 +1,226 @@
...
+
+  def getOutputOI(inputOI : ObjectInspector) : ObjectInspector = {
+    if (outputOI == null) {
+      outputOI = codegen.initialize(inputOI)
+    }
+
+    outputOI
+  }
+
+  def evaluator() = evaluateClass(packageName + "." + className, getCode(codegen, packageName, className)).newInstance()
```
rxin: Put the return type (IEvaluator?) here. And wrap the line so it is 100 col wide.

src/main/scala/shark/execution/cg/CodeGenManager.scala
```diff
@@ -0,0 +1,226 @@
...
+    val compiler = ToolProvider.getSystemJavaCompiler()
+    val dc = new DiagnosticCollector[JavaFileObject]()
+    var fileManager = new ClassFileManager(compiler
+      .getStandardFileManager(null, null, null))
+    var file : JavaFileObject = new JavaSourceObject(classFullName, code)
+    var compilationUnits = Arrays.asList(file)
+    var task = compiler.getTask(null, fileManager, dc, null, null, compilationUnits)
+    var result = task.call()
+
+    for (d <- dc.getDiagnostics()) {
```
rxin: Do you know in what cases would we get errors but succeed in compilation here?

If they are simply compilation errors, why don't we put them into the "else" branch later?

src/main/scala/shark/execution/cg/CodeGenManager.scala
```diff
@@ -0,0 +1,226 @@
...
+    outStream().println(code)
+
+    if (result) {
+      outStream().println("Success in Compiling: " + classFullName)
+      // TODO need to think about how to share the class among the cluster
+      fileManager.getClassLoader(null).loadClass(classFullName).asInstanceOf[Class[IEvaluate]]
+    } else {
+      var err = "Error in compiling class: " + classFullName
+      logError(err)
+      throw new Exception(err)
```
rxin: let's create a new exception type CodeGenException maybe? And use that throughout the code

src/main/scala/shark/execution/cg/IEvaluate.scala
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package shark.execution.cg
+
+import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector
+import org.apache.hadoop.hive.ql.exec.ExprNodeGenericFuncEvaluator
+
+abstract class IEvaluate {
+  def init(oi : ObjectInspector)
+  def evaluate(rowInspector : Object) : Object
```
rxin: The argument should be a "row" rather than a row inspector, isn't it?

src/main/scala/shark/execution/cg/ExprNodeEvaluatorWrapper.scala
```diff
@@ -0,0 +1,70 @@
...
+  override def initialize(inputOI : ObjectInspector) : ObjectInspector = {
+    var outputOI : ObjectInspector = null
+
+    if (inputOI.isInstanceOf[StructObjectInspector]) {
+      // not support the map/list. currently
+      try {
+        outputOI = this.cgm.getOutputOI(inputOI)
+        if (outputOI != null) {
+          var eval = cgm.evaluator()
+          eval.init(inputOI)
```
rxin: Any reason you want to wrap this around a CGExecutor and not use eval directly?

src/test/scala/shark/execution/cg/TestExprNodeCodeGen.scala
```diff
@@ -0,0 +1,1054 @@
...
+import org.apache.hadoop.hive.serde2.`lazy`.LazyFactory
+import org.apache.hadoop.hive.serde2.`lazy`.LazyStruct
+import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory
+import org.apache.hadoop.io.BooleanWritable
+import org.apache.hadoop.io.IntWritable
+import org.apache.hadoop.io.LongWritable
+import org.apache.hadoop.io.Text
+import org.junit.Before
```
rxin: If it is not too late to change, try use scalatest rather than junit. I find that sbt and the scala plugins tend to integrate better with scalatest in terms of error reporting. 

If you want to write the test in Java, I think JUnit is the perfect way to go.

rxin:
Chenghao,

I started looking at this tonight and only had time to cover 4 or 5 files today. Overall, this is nicely done - I like that it is very well designed and isolated changes. I will continue the review tomorrow and on Thursday.

I did notice a few style things. The project mostly follows the Scala style guide at http://docs.scala-lang.org/style/

A few things in particular:
1. Explicitly specify return types in function definitions, unless they are very very obvious (i.e. "def myFunc = 1" is obvious enough). That reduces the chance of programming error, and makes it easier to understand the code. I had to jump a few places to figure out the return type of a lot of functions.
2. Keep line width to 100 characters (with the exception for import statements which can really go as long as you want)
3. Add a space after ":", but no space before. i.e.  "def myfunc(variableName: VarType): ReturnType = { ... }"
4. Add a space after ","
5. Space before and after binary operators like "==", "=", ">=".

Thanks!

rxin:
Oh because of the complexity of this change, if you can add more inline comments to explain what certain functions/classes do, that would be great. (You can skip the self-evident ones)

src/main/scala/shark/execution/cg/CodeGenManager.scala
```diff
@@ -0,0 +1,226 @@
...
+import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector
+import org.apache.hadoop.hive.ql.plan.ExprNodeDesc
+
+import scala.collection.mutable.LinkedHashSet
+import scala.collection.JavaConversions._
+
+import shark.execution.cg.node.ExprNodeCodeGen
+import shark.LogHelper
+
+class Definition(
```
rxin: Should this be named "ClassDefinition" or "MethodDefinition"?

chenghao-intel:
Thanks rxin. I will update/refactor the code  accordingly.

src/main/scala/shark/execution/cg/ExprNodeEvaluatorWrapper.scala
chenghao-intel: Within the ExprNodeEvaluatorWrapper.evaluate(row :AnyRef), either using a flag variable or abstract class reference, to call the actual executor, Hive or Code Gen evaluator. I just picked up the later, to eliminate the flag checking for every row evaluating, and hope the JVM could optimize(inline) the virtual function calls.

Sorry, I should put more comments in the code.

src/test/scala/shark/execution/cg/TestExprNodeCodeGen.scala
chenghao-intel: Very good suggestion, I would like to try the scalatest.

rxin:
Thanks. Let me know when you update this. Git doesn't automatically notify me when you just add new commits - so leaving a comment here would be great.

chenghao-intel:
Sorry, I didn't know that. I am working on the code-refactor, and adding more comments. Will send out another update soon.

rxin:
Closing this one because there is a new pull request #101

PR #95, opened 2013-05-14
Fix for SHARK-138

sun-rui:


rxin:
Thanks for doing this. Do you have a test query that would produce the problem? It would be great to add it to the SQLSuite.scala.

sun-rui:
Add a test case.

src/test/scala/shark/SQLSuite.scala
```diff
@@ -73,6 +73,14 @@ class SQLSuite extends FunSuite with BeforeAndAfterAll {
       OVERWRITE INTO TABLE users""")
     sc.sql("drop table if exists users_cached")
     sc.sql("create table users_cached as select * from users")
+
+    // test1
+    sc.sql("drop table if exists test1")
+    sc.sql("""CREATE TABLE test1 (id INT, test1val ARRAY<INT>)
+      row format delimited fields terminated by '\t'""")
+    sc.sql("LOAD DATA LOCAL INPATH '${hiveconf:shark.test.data.path}/test1.txt' INTO TABLE test1")
```
rxin, line 81: test1.txt is missing here...

rxin:
Sorry for the delay in getting to this. Looks like one test file is missing.

BTW - is there any performance degradation here? It might be better to put an if check outside the while loop to see if there is any need in data conversion. If not, we can simply call the old while loop. If yes, we can use the copy option.

Thanks.

sun-rui:
test1.txt is not missing. It is located at data/files/test1.txt.

sun-rui:
For the copy operation, we just do the same as HIVE. Look at the following piece of code from computeValues() in JoinUtil.java:
    for (int i = 0; i < valueFields.size(); i++) {
      nr.add(ObjectInspectorUtils.copyToStandardObject(valueFields.get(i)
          .evaluate(row), valueFieldsOI.get(i),
          ObjectInspectorCopyOption.WRITABLE));
    }

I am not sure if the optimization that performs copy only when the result of a field evaluation is not standard object is proper here. Because firstly a recursive type check is needed when a field is a complex type, secondly maybe the copy operations are intended so that independent changes can be made later.

I would prefer we keep the change here and can investigate it if any performance issue is found later.

rxin:
Thanks I've resolved the conflicts and merged this.

PR #96, opened 2013-05-16
ENSIME + fixes for tests

junkumar:
Works with new sbt

### DIFF
--- a/src/test/scala/shark/SharkServerSuite.scala
+++ b/src/test/scala/shark/SharkServerSuite.scala
@@ -20,6 +20,9 @@ class SharkServerSuite extends FunSuite with BeforeAndAfterAll with ShouldMatche
   val METASTORE_PATH = TestUtils.getMetastorePath("server")
   val DRIVER_NAME  = "org.apache.hadoop.hive.jdbc.HiveDriver"
   val TABLE = "test"
+  // use a different port, than the hive standard 10000,
+  // for tests to avoid issues with the port being taken on some machines
+  val PORT = "9011"
 
   Class.forName(DRIVER_NAME)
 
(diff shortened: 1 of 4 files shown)


src/test/scala/shark/SharkServerSuite.scala
```diff
@@ -32,10 +35,15 @@ class SharkServerSuite extends FunSuite with BeforeAndAfterAll with ShouldMatche
...
-      "javax.jdo.option.ConnectionURL=jdbc:derby:;databaseName=" + METASTORE_PATH + ";create=true",
-      "-hiveconf",
-      "hive.metastore.warehouse.dir=" + WAREHOUSE_PATH)
+      "--verbose",
+      "-p",
+      PORT,
+      "--hiveconf",
+      "hive.root.logger=INFO,console",
+      "--hiveconf",
+      "\"javax.jdo.option.ConnectionURL=jdbc:derby:;databaseName=" + METASTORE_PATH + ";create=true\"",
```
rxin, line 44: Don't do the two level quote. Those quotes are only required in bash. When the process is launched via process builder, the entire string is realdy a single unit and  ";" doesn't separate the commands.

rxin:
Thanks. This looks good. Just a small change required.

src/test/scala/shark/SharkServerSuite.scala
junkumar, line 44: Why not leave the 2 level quote in the code?
It works correctly when launched via process builder.
It also has the added advantage of dumping something to the console that can be copied & pasted directly into another shell while debugging when you are not going to remember or notice the quotes.

rxin, line 44: ok the 2nd point convinced me.

PR #97, opened 2013-05-23
Shark 141

haitaoyao:
no lock for XmlSerializer with the patch on HIVE.
should be used with this pull request merged: 
https://github.com/amplab/hive/pull/19


rxin:
Unfortunately after a lot of investigation, I am afraid the xml decoder itself is not thread safe (it's not just the object construction). We plan to do some major refactoring of the Shark backend codebase, and will solve this problem in that refactoring. Closing this one now. Thanks a lot for the contribution.

PR #98, opened 2013-06-05
Shark-146 map join fails due to task result bigger than Akka frame size

shimingfei:
creating blocks for small table and put them in block manager on slave workers, use block manager to fetch the blocks

### DIFF
--- a/src/main/scala/shark/execution/MapJoinOperator.scala
+++ b/src/main/scala/shark/execution/MapJoinOperator.scala
@@ -35,7 +35,7 @@ import shark.SharkEnvSlave
 import shark.execution.serialization.{OperatorSerializationWrapper, SerializableWritable}
 import spark.{RDD, SparkEnv}
 import spark.broadcast.Broadcast
-
+import spark.storage.StorageLevel
 
 /**
  * A join operator optimized for joining a large table with a number of small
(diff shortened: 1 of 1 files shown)


rxin:
Does this actually work? It looks like you haven't added the blocks to the block manager on workers yet.

shimingfei:
yes it works in my test.
I used rddForHash.foreach(_=>Unit) to cache the rdd into memory.
then get it from workers

2013/6/5 Reynold Xin notifications@github.com

> Does this actually work? It looks like you haven't added the blocks to the
> block manager on workers yet.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/amplab/shark/pull/98#issuecomment-18960166
> .

## 

Thanks,
*Mingfei Shi
Intel Asia-Pacific R&D Ltd.
*mingfei.shi@intel.com

rxin:
Where is the code for that foreach?

shimingfei:
sorry, I made some mistake in the pull request, i will update it

2013/6/5 Reynold Xin notifications@github.com

> Where is the code for that foreach?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/amplab/shark/pull/98#issuecomment-18960535
> .

## 

Thanks,
*Mingfei Shi
Intel Asia-Pacific R&D Ltd.
*mingfei.shi@intel.com

JoshRosen:
I have an experimental Spark branch to collect results through the BlockManager (actually, it supports the existing mechanism for sending small results, too); see https://github.com/JoshRosen/spark/commits/results-through-blockmanager.  Also, there's some relevant discussion in https://github.com/mesos/spark/pull/610

src/main/scala/shark/execution/MapJoinOperator.scala
```diff
@@ -126,7 +126,19 @@ class MapJoinOperator extends CommonJoinOperator[MapJoinDesc, HiveMapJoinOperato
 
       // Collect the RDD and build a hash table.
       val startCollect = System.currentTimeMillis()
-      val wrappedRows: Array[(Seq[AnyRef], Seq[Array[AnyRef]])] = rddForHash.collect()
+      if (rddForHash.getStorageLevel == StorageLevel.NONE) {
+        rddForHash.persist()
```
rxin: Can you change persist to specify the memory and disk option? Otherwise if the block doesn't fit in memory, it just gets dropped and you won't be able to find it later.

Also, once the master fetches the block, it will need to unpersist the RDD.

rxin:
Thanks, Mingfei!

PR #99, opened 2013-06-11
Modify memoryBytesToString to use pattern matching, lazy contants, and d...

JasonGiedymin:
...isplays ranges up front.
- Modify memoryBytesToString to use pattern matching, lazy contants, and displays ranges up front.


rxin:
Thanks, Jason. Not sure if this particular code path matters a lot, but thanks!

JasonGiedymin:
Sure I know it isn't a critical path but I'm looking around trying to learn what's here. Maybe poke around shine some parts up just a little while I'm at it.

-Jason

PR #100, opened 2013-06-19
add copartition join to Shark

shimingfei:
Port from old Shark code, made some changes:
1.check whether join on partition columns or not
2.needn't use distribute by when cache tables.
eg:
set shark.copartition.partnum=128;
CREATE TABLE sales_cached TBLPROPERTIES("shark.partition.col"="_col3") AS select \* from sales;
CREATE TABLE item_cached TBLPROPERTIES("shark.copartition.table"="sales_cached", "shark.partition.col"="_col0") AS select \* from item;


rxin:
Thanks - I will take a look at this this week.

src/main/scala/shark/CoPartitioner.scala
```diff
@@ -0,0 +1,15 @@
+package shark
```
rxin: Probably best to put this in the execution package.

src/main/scala/shark/CoPartitioner.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import scala.collection.mutable.ArrayBuffer
+import spark.HashPartitioner
+
+class CoPartitioner(partitions: Int) extends HashPartitioner(partitions) {
+  
+  private val partCols = new ArrayBuffer[String]
```
rxin: Actually - ignore my previous comment.

Let's rename partCols to _partCols, and then rename the defPartCols function to partitionColumns

src/main/scala/shark/CoPartitioner.scala
```diff
@@ -0,0 +1,15 @@
...
+  
+  def addPartitionCols(col: String) {
```
rxin: This function should be named addPartitionColumn (not plural) since you are adding only one column.

src/main/scala/shark/execution/JoinOperator.scala
```diff
@@ -82,6 +82,18 @@ class JoinOperator extends CommonJoinOperator[JoinDesc, HiveJoinOperator]
     combineMultipleRdds(inputRdds)
   }
 
+  def rddsCorpartitioned(rdds: Seq[RDD[(_,_)]]): Boolean = {
+    if (rdds.size < 2)
+      return false
+    val partitioner = rdds(0).partitioner
+    if (None == partitioner)
```
rxin: Try to avoid using return as part of the control flow.

You can rewrite this as

```
if (rdds.size > 1 && rdds.head.partitioner != None) {
  val part = rdds.head.partitioner
  rdds.find(_.partitioner != part).isDefined
} else {
  false
}
```

src/main/scala/shark/execution/JoinOperator.scala
```diff
@@ -101,10 +113,13 @@ class JoinOperator extends CommonJoinOperator[JoinDesc, HiveJoinOperator]
     val rddsInJoinOrder = order.map { inputIndex =>
       rddsJavaMap.get(inputIndex.byteValue.toInt).asInstanceOf[RDD[(ReduceKey, Any)]]
     }
+    var partNum: Int = numReduceTasks
```
rxin: Let's make this a val.  You can do

```
val partNum: Int = {
  if (...) {
    rdds(0)._2.asInstanceOf[RDD[(_, _)]].partitioner.get.numPartitions
  } else {
    numReduceTasks
  }
}
```

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -61,6 +67,14 @@ class MemoryStoreSinkOperator extends TerminalOperator {
 
     val statsAcc = SharkEnv.sc.accumulableCollection(ArrayBuffer[(Int, TablePartitionStats)]())
     val op = OperatorSerializationWrapper(this)
+    val shouldPartition = partitionCol != ""
+ 
+    def getCacheLocations(rdd: RDD[_]) = {
```
rxin: Put the return type here.

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -70,18 +84,71 @@ class MemoryStoreSinkOperator extends TerminalOperator {
         null
       }
 
+    val rddToCache = {
+      if (shouldPartition) {
+        val (partitioner, locations) = {
+          val partNum = localHconf.getInt("shark.copartition.partnum", 5)
+          SharkEnv.memoryMetadataManager.get(coPartitionTableName) match {
+            case Some(coRdd) => logInfo("table found in cache:" + coPartitionTableName)
```
rxin: Perhaps make this logging message a little bit more useful - say something along the line: copartitioning table a with table b

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -70,18 +84,71 @@ class MemoryStoreSinkOperator extends TerminalOperator {
...
+            val value = serde.serialize(row, op.objectInspector).asInstanceOf[BytesWritable]
+            val valueArr = new Array[Byte](value.getLength)
+            Array.copy(value.getBytes, 0, valueArr, 0, valueArr.getLength)
+            (key, valueArr)
+          }
+        }
+        logInfo("kvRdd generated for table with copartitionCol " + partitionCol + " " + partitioner)
+        new ShuffledWithLocationsRDD(kvRdd, partitioner, locations).mapPartitions(
+            {iter => iter.map(x => x._2)}, true)
+      }else{
```
rxin: add extra space before else and after else

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -92,7 +159,7 @@ class MemoryStoreSinkOperator extends TerminalOperator {
         statsAcc += Tuple2(partitionIndex, new TablePartitionStats(Array(), 0))
         Iterator(new TablePartition(0, Array()))
       }
-    }
+    }, true)
```
rxin: for boolean values, use named arguments so it is easy to tell what "false" mean here.

i.e.

```
}, preservePartitioning = true)
```

src/main/scala/shark/execution/Operator.scala
```diff
@@ -33,6 +33,7 @@ import spark.RDD
 
 abstract class Operator[T <: HiveOperator] extends LogHelper with Serializable {
 
+  def preservesPartitioning = false
```
rxin, line 36: add a blank line

src/main/scala/shark/execution/ReduceSinkOperator.scala
```diff
@@ -73,6 +76,57 @@ class ReduceSinkOperator extends UnaryOperator[HiveReduceSinkOperator] {
     initializeOisAndSers(conf, objectInspector)
   }
 
+  def getKeyColsInfo = {
```
rxin: put return type here

src/main/scala/shark/execution/ReduceSinkOperator.scala
```diff
@@ -73,6 +76,57 @@ class ReduceSinkOperator extends UnaryOperator[HiveReduceSinkOperator] {
     initializeOisAndSers(conf, objectInspector)
   }
 
+  def getKeyColsInfo = {  
+    val keyColsName = new ArrayBuffer[String]
+    conf.getKeyCols.foreach { expr =>
```
rxin: you don't need the extra match here. can simply do

```
conf.getKeyCols.foreach {
  case coumnDesc: ExprNodeColumnDesc => ...
  case genericFuncDesc: ExprNodeGenericFuncDesc => ....
}
```

src/main/scala/shark/execution/ReduceSinkOperator.scala
```diff
@@ -73,6 +76,57 @@ class ReduceSinkOperator extends UnaryOperator[HiveReduceSinkOperator] {
...
+    keyColsName
+  }
+  
+  def getKeyColInternalName = {
+    val keyColsName = getKeyColsInfo
+    val tableColsNameToIndex = objectInspector.asInstanceOf[StructObjectInspector].
+          getAllStructFieldRefs.zipWithIndex.map { case (field, index) =>
+          (field.getFieldName() -> index)
+        }.toMap
+    keyColsName.map { keyCol =>
```
rxin: similar to above - you can simplify the pattern matching here

src/main/scala/shark/execution/ReduceSinkOperator.scala
```diff
@@ -73,6 +76,57 @@ class ReduceSinkOperator extends UnaryOperator[HiveReduceSinkOperator] {
...
+            case Some(pos) => "_col" + pos
+            case None => logError("Column not found in table:" + keyCol)
+            keyCol
+          }
+      }
+    }
+  }
+  
+  override def preprocessRdd(rdd: RDD[_]) = {
+    rdd.partitioner.foreach { part => part match {
```
rxin, line 131: again - simplify the pattern matching

src/main/scala/shark/execution/ReduceSinkOperator.scala
```diff
@@ -209,9 +263,13 @@ class ReduceSinkOperator extends UnaryOperator[HiveReduceSinkOperator] {
...
       val value = valueSer.serialize(evaluatedValue, valObjInspector).asInstanceOf[BytesWritable]
 
+      val valArr = new Array[Byte](value.getLength)
+      Array.copy(value.getBytes, 0, valArr, 0, value.getLength)
+      val valueCopy = new BytesWritable(valArr)
+ 
       reduceKey.bytesWritable = key
       reduceKey.partitionCode = partitionCode
-      (reduceKey, value)
+      (reduceKey.createDeepCopy, valueCopy)
```
rxin: We shouldn't create the deep copy here - because this will create tons of objects that are not needed on the map side. 

I think an easier way to solve this problem is outlined in the earlier email I sent you. If the table have the same partitioners, we should convert all ReduceKeyMapSide to ReduceKeyReduceSide (by copying the byte array). This way it simplifies your join code a little bit (no more type checking in the while loop), and also makes the compare function in ReduceKey easier to implement.

src/main/scala/shark/execution/ShuffledWithLocationsRDD.scala
```diff
@@ -0,0 +1,22 @@
...
+ * @param part the partitioner used to partition the RDD
+ * @tparam K the key class.
+ * @tparam V the value class.
+ */
+class ShuffledWithLocationsRDD[K, V](
+    @transient parent: RDD[(K, V)],
+    part: Partitioner,
+    locations: Array[Seq[String]]) extends ShuffledRDD(parent, part) {
+
+  override def getPreferredLocations(split: Partition) =
```
rxin: add return type, and wrap the whole function with { }

src/main/scala/shark/execution/ShuffledWithLocationsRDD.scala
```diff
@@ -0,0 +1,22 @@
...
+ * The resulting RDD from a shuffle (e.g. repartitioning of data).
+ * @param parent the parent RDD.
+ * @param part the partitioner used to partition the RDD
+ * @tparam K the key class.
+ * @tparam V the value class.
+ */
+class ShuffledWithLocationsRDD[K, V](
+    @transient parent: RDD[(K, V)],
+    part: Partitioner,
+    locations: Array[Seq[String]]) extends ShuffledRDD(parent, part) {
```
rxin: Mark locations as @transient

src/main/scala/shark/memstore2/MemoryMetadataManager.scala
```diff
@@ -41,7 +41,11 @@ class MemoryMetadataManager {
     _keyToRdd(key.toLowerCase) = rdd
   }
 
-  def get(key: String): Option[RDD[_]] = _keyToRdd.get(key.toLowerCase)
+  def get(key: String): Option[RDD[_]] =
```
rxin, line 45: maybe turn the thing into a single line

```
if (null == key) None else _keyToRdd.get(key.toLowerCase)
```

src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -177,14 +177,14 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
...
-    prunedRdd.mapPartitions { iter =>
+    prunedRdd.mapPartitions ({ iter =>
       if (iter.hasNext) {
         val tablePartition = iter.next.asInstanceOf[TablePartition]
         tablePartition.iterator
       } else {
         Iterator()
       }
-    }
+    }, true)
```
rxin: It is actually problematic in the map pruning case. If two tables are co-partitioned, and map pruning pruned some partitions of one of the tables, technically we cannot perform the join anymore! This is pretty tricky to solve. Can you think about this a little bit more?

rxin:
Hi Mingfei - this looks good overall. Thanks a lot for porting the code and taking care of some nitty gritty things.

I made some comments, mostly in style. But also there are two things that stand out:
1. We shouldn't make a deep copy of the ReduceKeyMapSide for the non-copartitioned-join case. I suggested a fix for that.
2. The current code path is very problematic for map pruning.
3. We should think about how to best test this, starting with some unit tests in SQLSuite.

Again, thanks for doing this! This will substantially improve the performance for joining two large tables.

shimingfei:
Hi Reynold,
I have modified the code according to your comments.

src/main/scala/shark/execution/Operator.scala
```diff
@@ -205,9 +207,9 @@ object Operator extends LogHelper {
    * to do logging, but calling logging automatically adds a reference to the
    * operator (which is not serializable by Java) in the Spark closure.
    */
-  def executeProcessPartition(operator: Operator[_ <: HiveOperator], rdd: RDD[_]): RDD[_] = {
+  def executeProcessPartition(operator: Operator[_ <: HiveOperator], rdd: RDD[_], preservesPartitioning: Boolean): RDD[_] = {
```
rxin: Can we wrap this line to fit it in 100 chars?

src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -51,7 +51,10 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
   @BeanProperty var firstConfPartDesc: PartitionDesc  = _
   @BeanProperty var tableDesc: TableDesc = _
   @BeanProperty var localHconf: HiveConf = _
-
+  var keepPartitioning = true
```
rxin: mark this as private.

src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -177,14 +180,18 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
         rdd
       }
 
-    prunedRdd.mapPartitions { iter =>
+    if(prunedRdd.partitions.size != rdd.partitions.size) {
```
rxin, line 190: Can you add an inline TODO here - in the future, we want to support copartitioned joins even for pruned rdds.

src/main/scala/shark/execution/serialization/HiveStructSerializer.scala
```diff
@@ -53,3 +54,24 @@ class HiveStructSerializer(val rowObjectInspector: StructObjectInspector) {
...
+    outputByteBuffer.reset()
+    for(i <- 0 until fields.size){
+      BinarySortableSerDe.serialize(outputByteBuffer, 
+         rowObjectInspector.getStructFieldData(row, fields(i)),
+         colObjectInspectors(i), 
+         invert)
+    }
+    val bytes = new Array[Byte](outputByteBuffer.length)
+    System.arraycopy(outputByteBuffer.getData(), 0, bytes, 0, outputByteBuffer.length)
+    new BytesWritable(bytes)
```
rxin: Would it work if we just return outputByteBuffer? That reduces the amount of temporary objects created.

src/main/scala/shark/execution/serialization/HiveStructSerializer.scala
```diff
@@ -53,3 +54,24 @@ class HiveStructSerializer(val rowObjectInspector: StructObjectInspector) {
...
+
+class HiveStructPartialSerializer(rowObjectInspector: StructObjectInspector, 
+                                  fields: Array[StructField]) {
+  private val outputByteBuffer = new OutputByteBuffer
+//  private val fields  = cols.map(rowObjectInspector.getStructFieldRef(_))
+  private val colObjectInspectors = fields.map(_.getFieldObjectInspector)
+  
+  def serialize(row: Object, invert: Boolean = false) = {
+    outputByteBuffer.reset()
+    for(i <- 0 until fields.size){
```
rxin: Change this to a while loop - also explicitly declare the return type for this function.

src/main/scala/shark/memstore2/MemoryMetadataManager.scala
```diff
@@ -35,20 +35,21 @@ class MemoryMetadataManager {
   private val _keyToStats: ConcurrentMap[String, collection.Map[Int, TablePartitionStats]] =
     new ConcurrentHashMap[String, collection.Map[Int, TablePartitionStats]]
 
-  def contains(key: String) = _keyToRdd.contains(key.toLowerCase)
+  def contains(key: String) = if(null == key) false else _keyToRdd.contains(key.toLowerCase)
```
rxin: Add whitespace after "if" (for all the if's in this file)

src/main/scala/shark/execution/ReduceSinkOperator.scala
```diff
@@ -73,6 +76,65 @@ class ReduceSinkOperator extends UnaryOperator[HiveReduceSinkOperator] {
     initializeOisAndSers(conf, objectInspector)
   }
 
+  def getPartColName: ArrayBuffer[String] = {
```
rxin: mark this as private - and can you add some comment explaining what this does? please do the same for getPartColInternalName

src/main/scala/shark/execution/CoGroupedRDD.scala
```diff
@@ -114,7 +114,23 @@ class CoGroupedRDD[K](@transient var rdds: Seq[RDD[(_, _)]], part: Partitioner)
     for ((dep, depNum) <- split.deps.zipWithIndex) dep match {
       case NarrowCoGroupSplitDep(rdd, itsSplitIndex, itsSplit) => {
         // Read them from the parent
-        for ((k, v) <- rdd.iterator(itsSplit, context)) { getSeq(k.asInstanceOf[K])(depNum) += v }
+        for ((k, v) <- rdd.iterator(itsSplit, context)) {
+          val k1 = k match {
```
rxin: Can we take the if/case match out of the for loop? It is better not to rely on branch prediction for speeding up the pattern matching performance.

src/main/scala/shark/execution/CoGroupedRDD.scala
```diff
@@ -114,7 +114,21 @@ class CoGroupedRDD[K](@transient var rdds: Seq[RDD[(_, _)]], part: Partitioner)
     for ((dep, depNum) <- split.deps.zipWithIndex) dep match {
       case NarrowCoGroupSplitDep(rdd, itsSplitIndex, itsSplit) => {
         // Read them from the parent
-        for ((k, v) <- rdd.iterator(itsSplit, context)) { getSeq(k.asInstanceOf[K])(depNum) += v }
+        rdd match {
+          case rdd: RDD[(ReduceKeyMapSide, BytesWritable)] =>
```
rxin, line 118: Would this match actually work? I am not sure if we keep enough type information to make this pattern match work (also consider how type erasure might affect this). One idea is you can put the block in getDependencies. Something like:

```
override def getDependencies: Seq[Dependency[_]] = {
  rdds.map { rdd =>
    if (rdd.partitioner == Some(part)) {
      logInfo("Adding one-to-one dependency with " + rdd)
      new OneToOneDependency(rdd.map { 
        // do the key, value clone to ReduceKeyReduceSide / Array[Byte] here
      })
    } else {
      logInfo("Adding shuffle dependency with " + rdd)
      new ShuffleDependency[Any, Any](rdd, part, SharkEnv.shuffleSerializerName)
    }
  }
}
```

shimingfei, line 118: I tried and that match actually works.  .

shimingfei:
Hi Reynold
I also change ShuffledWithLocationsRDD to use Shark's ShuffleSerializaer when doing Shuffle, and keep the return type of HiveStructPartialSerializer as BytesWritable

rxin:
Hi Mingfei,

I was running through the Hive unit tests, and saw a number of failures. Can you take a look and update the patch?

For example, groupby3.q gets the following exception:

```
F.Begin query: groupby3.q
scala.MatchError: Const int 5 (of class org.apache.hadoop.hive.ql.plan.ExprNodeConstantDesc)
    at shark.execution.ReduceSinkOperator$$anonfun$getGenericFuncColumnName$1$1.apply(ReduceSinkOperator.scala:84)
    at shark.execution.ReduceSinkOperator$$anonfun$getGenericFuncColumnName$1$1.apply(ReduceSinkOperator.scala:83)
    at scala.collection.Iterator$class.foreach(Iterator.scala:772)
    at scala.collection.JavaConversions$JIteratorWrapper.foreach(JavaConversions.scala:573)
    at scala.collection.IterableLike$class.foreach(IterableLike.scala:73)
    at scala.collection.JavaConversions$JListWrapper.foreach(JavaConversions.scala:615)
    at shark.execution.ReduceSinkOperator.getGenericFuncColumnName$1(ReduceSinkOperator.scala:83)
    at shark.execution.ReduceSinkOperator$$anonfun$getPartColName$1.apply(ReduceSinkOperator.scala:99)
    at shark.execution.ReduceSinkOperator$$anonfun$getPartColName$1.apply(ReduceSinkOperator.scala:94)
    at scala.collection.Iterator$class.foreach(Iterator.scala:772)
    at scala.collection.JavaConversions$JIteratorWrapper.foreach(JavaConversions.scala:573)
    at scala.collection.IterableLike$class.foreach(IterableLike.scala:73)
    at scala.collection.JavaConversions$JListWrapper.foreach(JavaConversions.scala:615)
    at shark.execution.ReduceSinkOperator.getPartColName(ReduceSinkOperator.scala:94)
    at shark.execution.ReduceSinkOperator.getPartColInternalName(ReduceSinkOperator.scala:107)
    at shark.execution.ReduceSinkOperator.preprocessRdd(ReduceSinkOperator.scala:127)
    at shark.execution.UnaryOperator.execute(Operator.scala:190)
    at shark.execution.Operator$$anonfun$executeParents$1.apply(Operator.scala:117)
    at shark.execution.Operator$$anonfun$executeParents$1.apply(Operator.scala:117)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:233)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:233)
    at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:60)
    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
```

If you have never run the Hive unit tests before, you can follow the guide: https://github.com/amplab/shark/wiki/Developer-Guide#hive-cli-tests

The following command runs through all the tests:

```
TEST_FILE=src/test/tests_pass.txt ./bin/dev/test
```

The following command runs only tests matching the name _groupby_:

```
TEST_FILE=src/test/tests_pass.txt TEST=groupby ./bin/dev/test
```

src/main/scala/shark/execution/ReduceSinkOperator.scala
```diff
@@ -73,6 +76,67 @@ class ReduceSinkOperator extends UnaryOperator[HiveReduceSinkOperator] {
...
+            keyCol
+          }
+      }
+    }
+  }
+  
+  override def preprocessRdd(rdd: RDD[_]) = {
+    val partColInternalName = getPartColInternalName
+    rdd.partitioner.foreach { part => part match {
+      case copart: CoPartitioner =>
```
rxin, line 132: Don't you also need to match for cases that are not a CoPartitioner? In the case of normal group by, we can see a HashPartitioner here also (and in the future maybe other Partitioner's).

shimingfei:
Hi Reynold, 
I will look into the issues and rerun the test cases before I push the code.

rxin:
@shimingfei any update on this?

shimingfei:
Hi Reynold,
I will update the patch later Today .

rxin:
Thanks, @shimingfei . Have you looked into how to use the actual column names instead of "_colx" yet?

shimingfei:
Hi Reynold,
Yes, I am working on it. After finishing the change, I will update the code

shimingfei:
Hi Reynold
I have finished the change, now the partition columns' names can be "_colx" or actual column names

rxin:
Excellent! Thanks.

I am going to try this out tomorrow and merge it. BTW, the pr can't be merged automatically anymore. Do you mind updating it?

shimingfei:
I will update it to latest shark code, if it works fine.

shimingfei:
Hi Reynold
I have merged the code to branch-0.8

PR #101, opened 2013-06-21
code gen for shark expression evaluating

chenghao-intel:
@rxin

This commit is based on the branch-0.8, and please ignore the previous pull request https://github.com/amplab/shark/pull/94.

Sorry, there are too much code, probably you'd better read from the unit test: shark.execution.cg.TestExprNodeCodeGen.

### DIFF
--- a/src/main/scala/shark/SharkConfVars.scala
+++ b/src/main/scala/shark/SharkConfVars.scala
@@ -52,6 +52,9 @@ object SharkConfVars {
 
   // If true, then query plans are compressed before being sent
   val COMPRESS_QUERY_PLAN = new ConfVar("shark.compressQueryPlan", true)
+  
+  // If true, will use the Code Gen expression evaluator
+  val EXPR_CG = new ConfVar("shark.expr.cg", true)
 
   // Add Shark configuration variables and their default values to the given conf,
   // so default values show up in 'set'.
@@ -70,6 +73,8 @@ object SharkConfVars {
       conf.setBoolean(MAP_PRUNING.varname, MAP_PRUNING.defaultBoolVal)
     if (conf.get(MAP_PRUNING_PRINT_DEBUG.varname) == null)
       conf.setBoolean(MAP_PRUNING_PRINT_DEBUG.varname, MAP_PRUNING_PRINT_DEBUG.defaultBoolVal)
+    if (conf.get(EXPR_CG.varname) == null)
+      conf.setBoolean(EXPR_CG.varname, EXPR_CG.defaultBoolVal)      
   }
 
   def getIntVar(conf: Configuration, variable: ConfVar): Int = {
--- a/src/main/scala/shark/execution/ExtractOperator.scala
+++ b/src/main/scala/shark/execution/ExtractOperator.scala
@@ -28,12 +28,16 @@ import org.apache.hadoop.io.BytesWritable
 
 import spark.RDD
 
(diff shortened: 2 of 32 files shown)


chenghao-intel:
@rxin 
The CodeGen expression evaluating does take the different behaviors with Hive in expression evaluating. 
The root reason of the differences is data casting from one type to another.
1) Exception raising V.S. null value retrieving
The utility class of PrimitiveObjectInspectorUtils, which used by Converter internally, may raise runtime exception on the fly, if the data types are not convertible; but the CodeGen computes types converting before the real data comes, constant null value will be provided if the data types are in-convertible.
2) Operator "<=>" 
In Hive, major steps computing for operator "<=>" are:
a. check if null for both left and right operand respectively. (apply the short cutting)
b. get the value for both left / right operand respectively (data converting if necessary, and exception may be thrown) (apply the short cutting)
c. check the equality for left and right operand
As in CodeGen, the step a and b is opposite (will try to make the data casting first), hence the short cutting behavior may not be the same, the unit test "TestExprNodeCodeGen" will print the full list of the different behavior cases.

I also checked the behavior of MySQL about the data type casting in expression evaluating:
1) it never raise exception while on the fly, even meet the in-convertible data type
2) null value will cause null value for the new data type
3) always try to provide the value for the data type casting, and default value (false or 0 etc.) provided for the in-convertible data types casting.

From my perspective, probably we may update the code for Hive (PrimitiveObjectInspectorUtils.java), to provide the default value, rather than raise exception in data type casting, and keep Hive Evaluating & CodeGen the same behavior in expression evaluating.

Actually, I created an jira issue (https://issues.apache.org/jira/browse/HIVE-4777) for discussion, but nobody replies.

What do you think?

Thanks, 
Cheng Hao

chenghao-intel:
@rxin 
And there are 3 other open issues in the code CGClassEntry.scala
1) In companion object CGClassEntry, it searches the cached NodeExprDesc and try to re-use CGClassEntry. this works in stand alone mode, but within the cluster env, "tuples.find(_._1.isSame(desc)) " always returns None. And I did some tests for NodeExprDesc, to comparing the desc instances before / after serialization / de-serialization via java's WriteObject API, which works fine, too. I am wondering if there any bugs of the shark customed  serialization / de-serialization API.
(Edit: Fixed see https://github.com/amplab/hive/pull/22)
2) Still in the companion object CGClassEntry, which try to cache the CGClassEntry for every NodeExprDesc, it may has the potential memory leak or performance issue in NodeExprDesc instance locating. I will make the cache as the LRU, the same for CacheClassLoader.
3) The JavaCompilerHelper works differently in sbt unit test,  as the JavaCompiler API always try to get the dependent jar from the AppClassLoader (which specified while java process launched via -cp option), as in sbt unit test, the -cp option takes only the "sbt-launch.jar", but in the real situation, the complete jars would be specified when the "StandaloneExecutorBackend" process starts. Hence the work around solution for sbt unit testing is, set the flag "FOR_UNIT_TEST_WORK_AROUND" in the unit test, and JavaCompilerHelper will try to load the dependent jars from the thread context classloader as the compilation option if the "FOR_UNIT_TEST_WORK_AROUND" was set. And there may be  bug for the customed UDF jars, which also loaded later on. Not sure if you has better solution for the dependent jar loading.

Thanks, 
Cheng Hao

chenghao-intel:
Update the file CGClassEntry.scala

PR #103, opened 2013-07-09
Compressing Columns Stored in Memory

junkumar:
This pull request provides functionality to compress columns stored in memory.

*\* Encoding schemes available
| _Scheme_                  | _Target column properties_                                     | _Column Types Currently Supported_ |
| Run Length Encoding(RLE)  | sorted OR small number of transitions OR Partition key columns | Int or String                      |
| Dictionary Encoding(Dict) | small number of unique values, but no runs                     | Int                                |
| Null Bit Vector Encoding  | sparse data, few non-null values                               | Int, Boolean, Long or Byte         |
| LZF Encoding              | slowest decompression speed  but best compression rate         | Int or String                      |
| NONE                      | raw storage; ideal if data has too many unique values          |                                    |

*\* Currently possible mixing of Encoding Schemes and column types
Nullable Column Types - {Byte, Short, Int, Long, Float, Double, Timestamp}

Null Encoding - Any Nullable Column Type
(Null Encoding is always optional when available)

RLE + Null - Int
RLE String

Dict + Null - Int

LZF - String

*\* How does it get used?
When a cached table is created in shark, all the data gets stored as columns
into memory. The compression is then used to store these columns based on their
properties.

*\* Compression results on production data at Yahoo
Tested on a 14 node cluster with a total of (14GB*14) 196 GB of RAM
Input data is a single table on HDFS/Hive with 4 parition keys and 3 other columns.

| _partition specification in input data_ | _number of rows_ | _HDFS bytes (RCfile compressed on disk)_  | _shark 0.7 bytes in memory (no compr)_ | _after compression bytes_  | _compr rate (% of uncompr)_ |
| action=ad_click p=1 dt=20130501         | 289,652          | 2,683,931     | 23,315,012                   | 8,832,576      |                       37.88 |
| p=1 dt=20130501                         | 153,494,679      | 964,635,051   | 13,602,326,763               | 2,565,180,073  |                       18.86 |
| action=search dt=20130501               | 226,388,853      | 7,350,168,969 | 18,066,874,698               | 6,353,214,544  |                       35.16 |
| action=search dt=2013050[1-5]           | 1,146,965,502    |               |                              | 91,496,897,244 |                             |

6 of the columns in this data were strings with no sort order.
RLE was chosen automatically whenever compression got used. Especially on the 4
partition columns.

*\* Uncompression results 
We are building simpler "light-weight" compression schemes instead of just using
established "heavy-weight" schemes like gzip or LZF so that we are not limited
by uncompression speeds. Heavy-weight schemes work well for data on disk because
disk is much slower than the CPU uncompression. Memory access speeds are much
faster and CPU uncompression can become a bottleneck.

This comparison measures uncompression speeds across multiple schemes. Input
data is a single slice of the data from the last test tested on a single node
setup. This is a single string column with data in sorted order, with a small
number of unique values; expect very high compression rates. These synthetic
benchmarks were simply used to check our implementation beliefs about the
bottleneck and do not numerically model how things will work in the production
clusters.

50 trial runs were executed (with the first 20 not counted) in each case to
measure averages.

A comparison of memory throughput versus data throughput is a useful way for us to reason about our implementations. The memory throughput for "None" compression will always be the best. But the data throughput will depend on how compressible the data is. The more that the data compresses the less space it will take and higher the data throughput will be for a fixed memory throughput.

[[https://dl.dropboxusercontent.com/u/7259597/SharkDecompressionSpeeds.jpg]]

| _scheme_ | _rows_     |    _ms_ | _raw MBps_ | _mem MBps_ | _% compr rate_ |
| None     | 2,500,000  |   74.87 |     413.99 |     541.38 |         100.00 |
| LZF      | 2,500,000  |   75.50 |     410.52 |      21.92 |           4.08 |
| RLE      | 2,500,000  |   34.56 |     896.65 |      62.61 |           5.34 |
| None     | 20,500,000 |  706.50 |     359.74 |     470.43 |         100.00 |
| LZF      | 20,500,000 |  765.30 |     332.10 |      17.73 |           4.08 |
| RLE      | 20,500,000 |  326.87 |     777.55 |      54.30 |           5.34 |
| None     | 90,500,000 | 3126.03 |     358.92 |     469.36 |         100.00 |
| LZF      | 90,500,000 | 3071.10 |     365.34 |      19.50 |           4.08 |
| RLE      | 90,500,000 | 1539.23 |     728.93 |      50.90 |           5.34 |

**\* Observations
- Zero compression (None) yields the highest raw memory throughput - the CPU is 
  interfering the least in this case.
- In a classic space-time trade-off, the fact that compressed data uses less
  memory allows nearly the same overall data throughput (raw MBps) even
  though the memory throughput (mem MBps) is lower due to uncompression.
- LZF has better compression rates for this data but cannot keep up with either
  no compression or RLE in term of uncompression speeds.
- RLE offers us the best compromise by allowing us to maintain the overall access
  speed while still letting us store more data into memory.
- Ultimately, like other columnar stores, we will be able to work directly with
  compressed data without needing uncompression which will make RLE's advantage
  even bigger.


rxin:
Thanks, Nandu!

Per our discussion offline, I am going to wait for Ram's pull request on column pruning and RDD dropping first since they are smaller changes. Hopefully we will merge all of them this week.

harsha2010:
Hi Nandu

There are a few issues with this pull request:
1. It includes a bunch of other commits that belong in separate pull requests (stats patch, column pruning etc). I am already in the process of creating pull requests for these patches and it should not all be conflated with the compression pull request.
2. It also seems to be pulling in yinst stuff which is yahoo specific
3. There is a bug in this patch where you are storing uniques references in column stats, and that can potentially cause the master node to end up storing data the size of the total RDD that is created, which is very problematic.
This can be fixed in one of two ways 
a) either move the uniques logic outside the stats code or
b) use hll sketch.
I prefer b) because hll sketches are configurable in size and can later be used to estimate unions/ intersections etc so have more information that just the unique counts.
Please look at the implementation of hll in the clearspring/ streamlib packages as to how to use these sketches.

rxin:
Another short-term solution for the uniques is to mark it as transient, and have an additional "numDistinct" field. In the long run, we should just move to hll.

junkumar:
Thanks for all the feedback!
The stats/bloomfilter commit was reverted before I sent the pull request. The column pruning is still in here but Reynold plans to merge this pr after your column pruning pr. So the end result should merge cleanly. 

Similarly any packaging/configuration stuff that is yahoo-specific was also not present in the final pull request.

I have moved the uniques IntArraySet to IntColumnBuilder so it does not get serialized. We can go to approximation estimates eventually.

harsha2010:
Nandu

I also see changes like the default serializer being configurable vs being hard coded as kryo serializer etc in this change set. Also, Hadoop version in shark build.scala is specific to our setup ... I don't want to go through every single commit in this change set but I think the correct way to create a pull request is to isolate ONLY your commits on compression

Ram

mbbui:
Sorry, if posting questions here is not appropriate, please suggest a better way.

@junkumar: I have some questions: 
1. Is the number aggregated over 14 nodes? If so can I say in None case, the raw bandwidth per node is 30MBps?
2. Why RLE produces higher raw bandwidth than None? Is it because your RLE encoder is run before shark tablepartition builder, and the combination of the two improve the overall bandwidth?
3. How did you calculate the memory bandwidth?

rxin:
Closing this one for @junkumar to submit another one that is more isolated.

ctn:
@michael-bbui 

> 1. Is the number aggregated over 14 nodes? If so can I say in None case, the raw bandwidth per node is 30MBps?

The "uncompression" perf test was done on a single node with highly-compressible synthetic data. The "raw throughput" is the uncompressed data size divided by how long it takes to uncompress it.

> 1. Why RLE produces higher raw bandwidth than None? Is it because your RLE encoder is run before shark tablepartition builder, and the combination of the two improve the overall bandwidth?

It just means that the test was able to process more equivalent raw data when compressed-uncompressed under the RLE encoding than None. In principle there is no limit to this "advantage". Imagine trying to read a file containing 10^20 zeroes of two formats, (a) actually containing 10^20 zeroes in sequence, and (b) an instruction saying "this file contains 10^20 zeroes". I'm sure you're already familiar with this, but I think perhaps you're reading more magic into the data table than is there. This doesn't have anything to do with the earlier phases, e.g., getting data from disk into a shark table, or even the compression cost itself.

> 1. How did you calculate the memory bandwidth?

That would be the number of compressed bytes processed by the uncompression scheme, divided by the uncompression time.

I _am_ curious about the apparent 30% adjustment in the `mem MBps` and `% comp rate`, though. Is that accounting for overhead or something, @junkumar ?

rxin:
@ctn as far as I understand the data  were real, not synthetic.

junkumar:
@ctn, good point. I will consider presenting this information in a different way to make that more obvious.
That 30% is overhead. On this particular string column, that was the typical overhead we saw for our data layout in memory as compared to just counting the number of bytes across all strings.
I was mainly measuring this to understand the associated cost.
In other words, if I had 10 strings and each had 3 characters for a total of 30 bytes. When I lay that out in memory it will cost approximately 40 bytes even without compression.
This is why you see see this change.

ctn:
Yes, I should have said it is "data with known properties, chosen for a particular benchmarking scenario, YMMV".

Related, the in-memory blowup factor is rather disconcerting, even after compression, and certainly before. I realize this is likely mostly due to Java. I wonder if there isn't anything we can do at the Shark or even JVM level to manage this, in addition to compression above Shark.

ctn:
@junkumar thanks for the very valuable measurements, btw.

rxin:
The 30% increase here for string columns is mostly from tracking the length of the string, which adds 4 bytes (a single int) per string, and this can be pretty big if the strings are short.

One simple way to reduce the overhead is to use zigzag encoding; the other is to figure out whether a column is fixed length, and if it is, don't track the length at all.

junkumar:
This request moved to https://github.com/amplab/shark/pull/108

PR #104, opened 2013-07-09
Support dropping tables 

harsha2010:
This patch adds the drop table feature to shark, while also fixing some issues with insert into cached tables not using the right storage level as the underlying cached tables.

### DIFF
--- a/src/main/scala/shark/RDDUtils.scala
+++ b/src/main/scala/shark/RDDUtils.scala
@@ -23,7 +23,9 @@ import scala.collection.JavaConverters._
 import com.google.common.collect.{Ordering => GOrdering}
 
 import spark.RDD
+import spark.rdd.UnionRDD
 import spark.SparkContext._
+import spark.storage.StorageLevel
 
 
 /**
@@ -31,6 +33,22 @@ import spark.SparkContext._
  * to Spark's built-in abstractions.
  */
 object RDDUtils {
+  
+  def getStorageLevelOfCachedTable(rdd: RDD[_]): StorageLevel = {
+    rdd match {
+      case u: UnionRDD[_] => u.rdds.foldLeft(rdd.getStorageLevel){
+        (s, r) => {
+          if (s == StorageLevel.NONE) {
+            getStorageLevelOfCachedTable(r)
+          } else {
+            s
+          }
+        }
+      }
+      case _ => rdd.getStorageLevel
+    }
+  }
+
   // TODO(rxin): Move this into execution pacakge.
 
   def sortLeastKByKey[K <% Ordered[K]: ClassManifest, V: ClassManifest](rdd: RDD[(K,V)], k: Int)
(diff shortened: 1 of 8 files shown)


src/main/scala/shark/memstore2/MemoryMetadataManager.scala
```diff
@@ -57,6 +58,26 @@ class MemoryMetadataManager {
   def getAllKeyStrings(): Seq[String] = {
     _keyToRdd.keys.collect { case k: String => k } toSeq
   }
+
+  def unpersist(key: String): Option[RDD[_]] = {
+    def unpersist(rdd: Option[RDD[_]]): Unit = {
```
rxin: Why does the argument take Option here instead of just RDD[_]?

src/main/scala/shark/parse/SharkDDLSemanticAnalyzer.scala
```diff
@@ -0,0 +1,23 @@
...
+import spark.RDD
+import spark.rdd.UnionRDD
+
+class SharkDDLSemanticAnalyzer(conf: HiveConf) extends DDLSemanticAnalyzer(conf) with LogHelper {
+
+  override def analyzeInternal(node: ASTNode): Unit = {
+    super.analyzeInternal(node)
+    //handle drop table query
+    if (node.getToken().getType() == HiveParser.TOK_DROPTABLE) {
+      SharkEnv.unpersist(getTableName(node))
```
rxin, line 15: Maybe check whether the table exists. If it doesn't throw an error when the table does not exist, it is fine too.

src/main/scala/shark/RDDUtils.scala
```diff
@@ -23,7 +23,9 @@ import scala.collection.JavaConverters._
 import com.google.common.collect.{Ordering => GOrdering}
 
 import spark.RDD
+import spark.rdd.UnionRDD
 import spark.SparkContext._
```
rxin, line 26: swap line 27 with line 26

rxin:
Thanks, Ram. This looks good overall. Just 3 small comments. In particular, one of them is about what happens if we are dropping a table that does not exist or is not cached.

src/main/scala/shark/parse/SharkDDLSemanticAnalyzer.scala
harsha2010, line 15: Hi

getTableName does not throw and error when the table does not exist. It
just parses the table name out of the sql.
Similarly, super.analyzeInternal does not throw an exception as well

rxin, line 15: I was referring to the unpersist method

harsha2010, line 15: yep, unpersist also works fine. I have made all the relevant changes as per
the code review. removed the option from the unpersist method as well to
make it more readable

Ram

PR #108, opened 2013-07-16
Compressing Columns Stored in Memory

junkumar:
Functionality to compress columns stored in memory (cached tables).
This is a cleaned-up version of https://github.com/amplab/shark/pull/103

To see a more friendly version of this document go to -
https://github.com/junkumar/shark/wiki/Compressing-Columns-Stored-in-Memory

*\* Intro
This document details the design of the compression options for data stored in cached tables in shark. The goals are to 
1. Allow shark to store more data in a fixed amount of memory without sacrificing read access rates 
2. Provide direct access to the compressed data so that query execution does not require uncompression.

*\* How does it get used?
When a cached table is created in shark, all the data gets stored as columns
into memory. Compression can then be used to store these columns more efficiently in memory. The default behavior is to determine whether compression is appropriate and which compression scheme should be used based on the properties of the data being stored in memory.

*\* Compression schemes available
| _Scheme_                  | _Target column properties_                                     | _Column Types Currently Supported_ |
| Run Length Encoding(RLE)  | sorted OR small number of transitions OR Partition key columns | Int or String                      |
| Dictionary Encoding(Dict) | small number of unique values, but no runs                     | Int                                |
| Null Bit Vector Encoding  | sparse data, few non-null values                               | Int, Boolean, Long or Byte         |
| LZF Encoding              | slowest decompression speed  but best compression rate         | Int or String                      |
| NONE                      | raw storage; ideal if data has too many unique values          |                                    |
| Auto  (Default)               | Shark chooses one of the above compression schemes that it deems ideal based on the properties of the column data     | All                                   |

Null Encoding can be mixed with one of Dict, RLE or LZF Encoding on Int columns

*\* Compression results on production data at Yahoo
Tested on a 14 node cluster with a total of (14GB*14) 196 GB of RAM
Input data is a single table on HDFS/Hive with 4 parition keys and 3 other columns.

| _partition specification in input data_ | _number of rows_ | _HDFS bytes (RCfile compressed on disk)_  | _shark 0.7 bytes in memory (no compr)_ | _after compression bytes_  | _compr rate (% of uncompr)_ |
| action=ad_click p=1 dt=20130501         | 289,652          | 2,683,931     | 23,315,012                   | 8,832,576      |                       37.88 |
| p=1 dt=20130501                         | 153,494,679      | 964,635,051   | 13,602,326,763               | 2,565,180,073  |                       18.86 |
| action=search dt=20130501               | 226,388,853      | 7,350,168,969 | 18,066,874,698               | 6,353,214,544  |                       35.16 |
| action=search dt=2013050[1-5]           | 1,146,965,502    |               |                              | 91,496,897,244 |                             |

6 of the columns in this data were strings with no sort order and the seventh column was a integer column.
RLE was chosen automatically whenever compression got used. Especially on the 4
partition columns.

*\* Uncompression results 
We are building simpler "light-weight" compression schemes instead of just using
established "heavy-weight" schemes like gzip or LZF so that we are not limited
by uncompression speeds. Heavy-weight schemes work well for data on disk because
disk is much slower than the CPU uncompression. Memory access speeds are much
faster and CPU uncompression can become a bottleneck.

This comparison measures uncompression speeds across multiple schemes. Input
data is a single slice of the data from the last test tested on a single node
setup. This is a single string column with data in sorted order, with a small
number of unique values; expect very high compression rates. These synthetic
benchmarks were simply used to check our implementation beliefs about the
bottleneck and do not numerically model how things will work in the production
clusters.

50 trial runs were executed (with the first 20 not counted) in each case to
measure averages.

A comparison of memory throughput versus data throughput is a useful way for us to reason about our implementations. The memory throughput for "None" compression will always be the best. But the data throughput will depend on how compressible the data is. The more that the data compresses the less space it will take and higher the data throughput will be for a fixed memory throughput.

[[https://dl.dropboxusercontent.com/u/7259597/SharkDecompressionSpeeds.jpg]]

| _scheme_ | _rows_     |    _ms_ |  _data MBps_ |  _mem MBps_ |   _% compr rate_ |
| None     | 2,500,000  |   74.87 |     413.99 |     541.38 |         100.00 |
| LZF      | 2,500,000  |   75.50 |     410.52 |      21.92 |           4.08 |
| RLE      | 2,500,000  |   34.56 |     896.65 |      62.61 |           5.34 |
| None     | 20,500,000 |  706.50 |     359.74 |     470.43 |         100.00 |
| LZF      | 20,500,000 |  765.30 |     332.10 |      17.73 |           4.08 |
| RLE      | 20,500,000 |  326.87 |     777.55 |      54.30 |           5.34 |
| None     | 90,500,000 | 3126.03 |     358.92 |     469.36 |         100.00 |
| LZF      | 90,500,000 | 3071.10 |     365.34 |      19.50 |           4.08 |
| RLE      | 90,500,000 | 1539.23 |     728.93 |      50.90 |           5.34 |

**\* Observations
- Zero compression (None) yields the highest raw memory throughput - the CPU is 
  interfering the least in this case.
- In a classic space-time trade-off, the fact that compressed data uses less
  memory allows nearly the same overall data throughput (raw MBps) even
  though the memory throughput (mem MBps) is lower due to uncompression.
- LZF has better compression rates for this data but cannot keep up with either
  no compression or RLE in term of uncompression speeds.
- RLE offers us the best compromise by allowing us to maintain the overall access
  speed while still letting us store more data into memory.
- Ultimately, like other columnar stores, we will be able to work directly with
  compressed data without needing uncompression which will make RLE's advantage
  even bigger.

*\* The cost of compression
Disk access times are expected to dominate the time taken to create cache tables. But the cost of compression can be observed as an increase in the time taken to create cache tables and must be traded off against the benefits of storing compressed data in memory.

*\* Specifying compression in SQL
We expect the "Auto" scheme to work well in most situations. However, this can be overridden in SQL when desired. Currently compression schemes can only be set for all columns of a given data type in a cached table.
=CREATE TABLE test_cached 
TBLPROPERTIES("shark.column.compress.int"="RLE") 
AS SELECT \* FROM test;=

=CREATE TABLE test_cached 
TBLPROPERTIES("shark.column.compress.string"="None") 
AS SELECT \* FROM test;=

### DIFF
--- a/src/main/scala/shark/SharkCliDriver.scala
+++ b/src/main/scala/shark/SharkCliDriver.scala
@@ -322,6 +322,7 @@ class SharkCliDriver(loadRdds: Boolean = false) extends CliDriver with LogHelper
           if (end > start) {
             val timeTaken:Double = (end - start) / 1000.0
             console.printInfo("Time taken: " + timeTaken + " seconds", null)
+            logInfo("Time taken: " + timeTaken + " seconds")
           }
 
           // Destroy the driver to release all the locks.
--- a/src/main/scala/shark/SharkCliDriver.scala
+++ b/src/main/scala/shark/SharkCliDriver.scala
@@ -322,7 +322,6 @@ class SharkCliDriver(loadRdds: Boolean = false) extends CliDriver with LogHelper
           if (end > start) {
             val timeTaken:Double = (end - start) / 1000.0
             console.printInfo("Time taken: " + timeTaken + " seconds", null)
-            logInfo("Time taken: " + timeTaken + " seconds")
           }
 
           // Destroy the driver to release all the locks.
--- a/src/main/scala/shark/SharkConfVars.scala
+++ b/src/main/scala/shark/SharkConfVars.scala
@@ -54,10 +54,8 @@ object SharkConfVars {
   val COMPRESS_QUERY_PLAN = new ConfVar("shark.compressQueryPlan", true)
 
   // Various compression schemes possible for storing data in memory
-  val COLUMNAR_COMPR_STRING = 
-    new ConfVar("shark.columnar.compr.string", "auto")
-  val COLUMNAR_COMPR_INT = 
-    new ConfVar("shark.columnar.compr.int", "auto")
+  val COLUMNAR_COMPR_STRING = new ConfVar("shark.columnar.compr.string", "auto")
+  val COLUMNAR_COMPR_INT = new ConfVar("shark.columnar.compr.int", "auto")
 
   // Add Shark configuration variables and their default values to the given conf,
   // so default values show up in 'set'.
@@ -74,10 +72,12 @@ object SharkConfVars {
       conf.setBoolean(COMPRESS_QUERY_PLAN.varname, COMPRESS_QUERY_PLAN.defaultBoolVal)
     if (conf.get(MAP_PRUNING.varname) == null)
       conf.setBoolean(MAP_PRUNING.varname, MAP_PRUNING.defaultBoolVal)
-    // if (conf.get(COLUMNAR_COMPR_INT.varname) == null)
-    //   conf.set(COLUMNAR_COMPR_INT.varname, COLUMNAR_COMPR_INT.defaultVal)
-    // if (conf.get(COLUMNAR_COMPR_STRING.varname) == null)
-    //   conf.set(COLUMNAR_COMPR_STRING.varname, COLUMNAR_COMPR_STRING.defaultVal)
+    if (conf.get(MAP_PRUNING_PRINT_DEBUG.varname) == null)
+      conf.setBoolean(MAP_PRUNING_PRINT_DEBUG.varname, MAP_PRUNING_PRINT_DEBUG.defaultBoolVal)
+    if (conf.get(COLUMNAR_COMPR_INT.varname) == null)
+      conf.set(COLUMNAR_COMPR_INT.varname, COLUMNAR_COMPR_INT.defaultVal)
+    if (conf.get(COLUMNAR_COMPR_STRING.varname) == null)
+      conf.set(COLUMNAR_COMPR_STRING.varname, COLUMNAR_COMPR_STRING.defaultVal)
   }
 
   def getIntVar(conf: Configuration, variable: ConfVar): Int = {
--- a/src/main/scala/shark/SharkConfVars.scala
+++ b/src/main/scala/shark/SharkConfVars.scala
@@ -53,6 +53,12 @@ object SharkConfVars {
   // If true, then query plans are compressed before being sent
   val COMPRESS_QUERY_PLAN = new ConfVar("shark.compressQueryPlan", true)
 
+  // Various compression schemes possible for storing data in memory
+  val COLUMNAR_COMPR_STRING = 
+    new ConfVar("shark.columnar.compr.string", "auto")
+  val COLUMNAR_COMPR_INT = 
+    new ConfVar("shark.columnar.compr.int", "auto")
+
   // Add Shark configuration variables and their default values to the given conf,
   // so default values show up in 'set'.
   def initializeWithDefaults(conf: Configuration) {
@@ -68,8 +74,10 @@ object SharkConfVars {
       conf.setBoolean(COMPRESS_QUERY_PLAN.varname, COMPRESS_QUERY_PLAN.defaultBoolVal)
     if (conf.get(MAP_PRUNING.varname) == null)
       conf.setBoolean(MAP_PRUNING.varname, MAP_PRUNING.defaultBoolVal)
-    if (conf.get(MAP_PRUNING_PRINT_DEBUG.varname) == null)
-      conf.setBoolean(MAP_PRUNING_PRINT_DEBUG.varname, MAP_PRUNING_PRINT_DEBUG.defaultBoolVal)
+    // if (conf.get(COLUMNAR_COMPR_INT.varname) == null)
+    //   conf.set(COLUMNAR_COMPR_INT.varname, COLUMNAR_COMPR_INT.defaultVal)
+    // if (conf.get(COLUMNAR_COMPR_STRING.varname) == null)
+    //   conf.set(COLUMNAR_COMPR_STRING.varname, COLUMNAR_COMPR_STRING.defaultVal)
   }
 
   def getIntVar(conf: Configuration, variable: ConfVar): Int = {
--- a/src/main/scala/shark/SharkConfVars.scala
+++ b/src/main/scala/shark/SharkConfVars.scala
@@ -54,8 +54,8 @@ object SharkConfVars {
   val COMPRESS_QUERY_PLAN = new ConfVar("shark.compressQueryPlan", true)
 
   // Various compression schemes possible for storing data in memory
-  val COLUMNAR_COMPR_STRING = new ConfVar("shark.columnar.compr.string", "auto")
-  val COLUMNAR_COMPR_INT = new ConfVar("shark.columnar.compr.int", "auto")
+  val COLUMN_COMPRESS_STRING = new ConfVar("shark.column.compress.string", "auto")
+  val COLUMN_COMPRESS_INT = new ConfVar("shark.column.compress.int", "auto")
 
   // Add Shark configuration variables and their default values to the given conf,
   // so default values show up in 'set'.
@@ -74,10 +74,10 @@ object SharkConfVars {
       conf.setBoolean(MAP_PRUNING.varname, MAP_PRUNING.defaultBoolVal)
     if (conf.get(MAP_PRUNING_PRINT_DEBUG.varname) == null)
       conf.setBoolean(MAP_PRUNING_PRINT_DEBUG.varname, MAP_PRUNING_PRINT_DEBUG.defaultBoolVal)
-    if (conf.get(COLUMNAR_COMPR_INT.varname) == null)
-      conf.set(COLUMNAR_COMPR_INT.varname, COLUMNAR_COMPR_INT.defaultVal)
-    if (conf.get(COLUMNAR_COMPR_STRING.varname) == null)
-      conf.set(COLUMNAR_COMPR_STRING.varname, COLUMNAR_COMPR_STRING.defaultVal)
+    if (conf.get(COLUMN_COMPRESS_INT.varname) == null)
+      conf.set(COLUMN_COMPRESS_INT.varname, COLUMN_COMPRESS_INT.defaultVal)
+    if (conf.get(COLUMN_COMPRESS_STRING.varname) == null)
+      conf.set(COLUMN_COMPRESS_STRING.varname, COLUMN_COMPRESS_STRING.defaultVal)
   }
 
   def getIntVar(conf: Configuration, variable: ConfVar): Int = {
--- a/src/main/scala/shark/SharkEnv.scala
+++ b/src/main/scala/shark/SharkEnv.scala
@@ -82,7 +82,6 @@ object SharkEnv extends LogHelper {
 
   System.setProperty("spark.serializer", classOf[spark.KryoSerializer].getName)
   System.setProperty("spark.kryo.registrator", classOf[KryoRegistrator].getName)
-  System.setProperty("spark.serializer", classOf[spark.KryoSerializer].getName)
 
   val executorEnvVars = new HashMap[String, String]
   executorEnvVars.put("SCALA_HOME", getEnv("SCALA_HOME"))
--- a/src/main/scala/shark/SharkEnv.scala
+++ b/src/main/scala/shark/SharkEnv.scala
@@ -36,7 +36,6 @@ object SharkEnv extends LogHelper {
           System.getenv("SPARK_HOME"),
           Nil,
           executorEnvVars)
-      sc.addSparkListener(new StatsReportListener())
     }
     sc
   }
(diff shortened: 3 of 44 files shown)


src/main/scala/shark/memstore2/column/StringColumnBuilder.scala
```diff
@@ -57,6 +61,7 @@ class StringColumnBuilder extends ColumnBuilder[Text] {
     _lengthArr.add(v.getLength)
     _arr.addElements(_arr.size, v.getBytes, 0, v.getLength)
     _stats.append(v)
+    _uniques += v
```
harsha2010: this looks like a bug to me: Hive and Shark reuse Text instances for reducing GC pressure (which is why the getLength method is provided on Text). If you have to compute uniques and store them in a set, you need to copy the Text, otherwise you are computing incorrect results.

src/main/scala/shark/memstore2/column/IntColumnBuilder.scala
```diff
@@ -47,33 +59,162 @@ class IntColumnBuilder extends ColumnBuilder[Int] {
...
+    val transitionsRatio = (_stats.transitions).toDouble / _nonNulls.size
+
+    if (selectivity > 0.2) {
+      CompressionScheme.None // compressed data might be bigger than original - don't bother trying
+    } else if (transitionsRatio < 0.5) {
+      RLE
+    } else if (uniques.size < MAX_DICT_UNIQUE_VALUES) {
+      Dict
+    } else {
+      LZF // low selectivity, but large number of uniques - LZF might offer good trade-off
```
harsha2010: You can potentially return LZF compression here, but below in the scheme matching, you are treating it as _ and throwing an exception. Either you should be handling LZF compression correctly in the scheme matching, or retuning None here

src/main/scala/shark/memstore2/column/IntColumnBuilder.scala
```diff
@@ -47,33 +59,162 @@ class IntColumnBuilder extends ColumnBuilder[Int] {
   }
 
   override def append(v: Int) {
-    _arr.add(v)
+    _nonNulls.add(v)
     _stats.append(v)
+    uniques.add(v)
```
harsha2010: 1. Keeping track of all the uniques is highly expensive
2. You don't need to track all uniques anyway since your MAX_DICT_UNIQUE_VALUES=256, you need only track upto 256 uniques and then give up on Dictionary Encoding.

src/main/scala/shark/SharkCliDriver.scala
```diff
@@ -322,6 +322,7 @@ class SharkCliDriver(loadRdds: Boolean = false) extends CliDriver with LogHelper
           if (end > start) {
             val timeTaken:Double = (end - start) / 1000.0
             console.printInfo("Time taken: " + timeTaken + " seconds", null)
+            logInfo("Time taken: " + timeTaken + " seconds")
```
rxin: remove this line

src/main/scala/shark/SharkConfVars.scala
```diff
@@ -53,6 +53,10 @@ object SharkConfVars {
   // If true, then query plans are compressed before being sent
   val COMPRESS_QUERY_PLAN = new ConfVar("shark.compressQueryPlan", true)
 
+  // Various compression schemes possible for storing data in memory
+  val COLUMNAR_COMPR_STRING = new ConfVar("shark.columnar.compr.string", "auto")
```
rxin: Let's rename the option to 

```
val COLUMN_COMPRESS_STRING = new ConfVar("shark.column.compress.string", "auto")
val COLUMN_COMPRESS_INT = new ConfVar("shark.column.compress.int", "auto")
```

src/main/scala/shark/SharkEnv.scala
```diff
@@ -82,6 +82,7 @@ object SharkEnv extends LogHelper {
 
   System.setProperty("spark.serializer", classOf[spark.KryoSerializer].getName)
   System.setProperty("spark.kryo.registrator", classOf[KryoRegistrator].getName)
+  System.setProperty("spark.serializer", classOf[spark.KryoSerializer].getName)
```
rxin: Remove this line

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -70,28 +75,40 @@ class MemoryStoreSinkOperator extends TerminalOperator {
...
+    var tableProperties = op.localHiveOp.getConf.getTableInfo.getProperties()
+    if (createTableProperties.get(SharkConfVars.COLUMNAR_COMPR_STRING.varname) != null) {
+      tableProperties.setProperty(SharkConfVars.COLUMNAR_COMPR_STRING.varname,
+        createTableProperties.get(SharkConfVars.COLUMNAR_COMPR_STRING.varname).toString)
+    }
+    if (createTableProperties.get(SharkConfVars.COLUMNAR_COMPR_INT.varname) != null) {
+      tableProperties.setProperty(SharkConfVars.COLUMNAR_COMPR_INT.varname,
+        createTableProperties.get(SharkConfVars.COLUMNAR_COMPR_INT.varname).toString)
+    }
+    op.logInfo("Op Table Properties - " + tableProperties)
```
rxin: Make a more meaningful log message "Op Table Properties". Perhaps "Creating table with properties: "?

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -70,28 +75,40 @@ class MemoryStoreSinkOperator extends TerminalOperator {
...
-      val serde = new ColumnarSerDe
-      serde.initialize(op.hconf, op.localHiveOp.getConf.getTableInfo.getProperties())
-
-      // Serialize each row into the builder object.
-      // ColumnarSerDe will return a TablePartitionBuilder.
-      var builder: Writable = null
-      iter.foreach { row =>
-        builder = serde.serialize(row.asInstanceOf[AnyRef], op.objectInspector)
-      }
+    var rdd: RDD[TablePartition] = inputRdd.mapPartitionsWithIndex {
```
rxin: change the style back. this block of code shouldn't have been changed.

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -101,12 +118,14 @@ class MemoryStoreSinkOperator extends TerminalOperator {
       SharkEnv.memoryMetadataManager.put(tableName, rdd)
 
       tachyonWriter.createTable(ByteBuffer.allocate(0))
-      rdd = rdd.mapPartitionsWithIndex { case(partitionIndex, iter) =>
-        val partition = iter.next()
-        partition.toTachyon.zipWithIndex.foreach { case(buf, column) =>
-          tachyonWriter.writeColumnPartition(column, partitionIndex, buf)
-        }
-        Iterator(partition)
+      rdd = rdd.mapPartitionsWithIndex {
```
rxin: change the style back too

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -167,8 +186,9 @@ class MemoryStoreSinkOperator extends TerminalOperator {
     }
 
     if (SharkConfVars.getBoolVar(localHconf, SharkConfVars.MAP_PRUNING_PRINT_DEBUG)) {
-      columnStats.foreach { case(index, tableStats) =>
-        println("Partition " + index + " " + tableStats.toString)
+      columnStats.foreach {
```
rxin: and here - style back

src/main/scala/shark/memstore2/column/LZFBlockColumnIterator.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  val baseIter: T = {
+    val ctor = baseIterCls.getConstructor(classOf[ByteBufferReader])
+    val uncompressedBBR = ByteBufferReader.createUnsafeReader(uncompressedBB)
+    ctor.newInstance(uncompressedBBR).asInstanceOf[T]
+  }
+
+
+  var rowCount = 0
+  override def next = {
```
rxin: use next() because next() has side effect.

src/main/scala/shark/memstore2/column/LZFColumnIterator.scala
```diff
@@ -0,0 +1,58 @@
...
+  val (num, compressedArr) = LZFSerializer.readFromBuffer(bytes)
+  val uncompressedArr = LZFSerializer.decode(compressedArr)
+  val uncompressedBytes = ByteBuffer.wrap(uncompressedArr)
+  val baseIter: T = {
+    val ctor = baseIterCls.getConstructor(classOf[ByteBufferReader])
+    val uncompressedBBR = ByteBufferReader.createUnsafeReader(uncompressedBytes)
+    ctor.newInstance(uncompressedBBR).asInstanceOf[T]
+  }
+
+  override def next = {
```
rxin: use next() because next() has side effect.

src/main/scala/shark/memstore2/column/Dictionary.scala
```diff
@@ -0,0 +1,125 @@
...
+  override def initialize(u: Array[Int]) = {
+    uniques = new Array[Int](u.size)
+    u.copyToArray(uniques)
+  }
+
+  // return the Int associated with Byte
+  override def get(b:Byte): Int = {
+    // Convert byte to index
+    // -128, 128 ===> 0, 256
+    var idx: Int = b
```
rxin: Can we write this slightly more functionally? Declare a val for this

```
val index = if (b < 0) b + 256 else b
uniques(index)
```

I am not sure if the IndexOutOfBoundsException exception is necessary, since accessing an array with an invalid index will throw a similar exception already.

src/main/scala/shark/memstore2/column/Dictionary.scala
```diff
@@ -0,0 +1,125 @@
...
+  override def getWritable(b: Byte): Object = {
+    writable.set(get(b))
+    writable
+  }
+
+  // return the Byte representation for Int value
+  override def getByte(i: Int): Byte = 
+    uniques.indexOf(i).toByte
+
+  override def sizeInBits = {
```
rxin: Let's change this to sizeInBytes ... so we don't need all those /8 in other places.

src/main/scala/shark/memstore2/column/Dictionary.scala
```diff
@@ -0,0 +1,125 @@
...
+// class StringDictionary extends Dictionary[Text]{
+
+// }
+
+/** Helper to share code betwen the Iterator and various Builders.
+  */
+object DictionarySerializer{
+
+  // Append the serialized bytes of the Dictionary into the ByteBuffer.
+  def writeToBuffer(buf: ByteBuffer, bitmap: IntDictionary) {
```
rxin: Why is this named "bitmap"?

src/main/scala/shark/memstore2/column/Dictionary.scala
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+
+// class StringDictionary extends Dictionary[Text]{
+
+// }
+
+/** Helper to share code betwen the Iterator and various Builders.
+  */
+object DictionarySerializer{
```
rxin: Either rename this to IntDictionarySerializer, or rename the methods writeToBuffer to writeIntDictToBuffer and readFromBuffer to readIntDictFromBuffer. 

Also add a space before {

src/main/scala/shark/memstore2/column/IntColumnBuilder.scala
```diff
@@ -47,33 +59,162 @@ class IntColumnBuilder extends ColumnBuilder[Int] {
...
+      "uniques=" + uniques.size + 
+      " selectivity=" + selectivity +
+      " transitionsRatio=" + transitionsRatio + 
+      " transitions=" + _stats.transitions +
+      " #values=" + _nonNulls.size)
+
+
+    scheme match {
+      case None => {
+        val minbufsize = (_nonNulls.size*4) + ColumnIterator.COLUMN_TYPE_LENGTH + sizeOfNullBitmap
```
rxin: rename the variable to just bufSize

src/main/scala/shark/memstore2/column/RLESerializer.scala
```diff
@@ -0,0 +1,172 @@
...
+
+import scala.collection.mutable._
+import scala.annotation.tailrec
+import shark.memstore2.buffer.ByteBufferReader
+
+import java.nio.{ ByteBuffer, IntBuffer, ByteOrder }
+import it.unimi.dsi.fastutil.ints.IntArrayList
+
+
+/* Decorator attempt with abstract override - incomplete - nandu
```
rxin: Let's remove this if it is no longer used.

src/main/scala/shark/memstore2/column/RLESerializer.scala
```diff
@@ -0,0 +1,172 @@
...
+      coded.toList
+    }
+  }
+
+  private var previous: A = fac()
+  private var previousRun: Int = 0
+
+  def encodeSingle(current: A): Unit =
+    if (previousRun > 0) {
+      if(previous == current) { // same value
```
rxin: How would this work for Text, in which == compares only if the Text's are the same object?

rxin:
I am going to do the merge manually.

rxin:
BTW this was merged together with the other pull request.

PR #110, opened 2013-07-26
drop table works for tachyon table.

rxin:


src/main/scala/shark/SharkEnv.scala
```diff
@@ -110,6 +110,17 @@ object SharkEnv extends LogHelper {
...
   def unpersist(key: String): Option[RDD[_]] = {
+    if (SharkEnv.tachyonUtil.tachyonEnabled()) {
+      if (SharkEnv.tachyonUtil.tableExists(key)) {
+        logInfo("Table " + key + " is in Tachyon.");
+        if (SharkEnv.tachyonUtil.dropTable(key)) {
+          logInfo("In Tachyon Table " + key + " was deleted.");
+        }
+      }
+    } else {
+      logInfo("Tachyon is not enabled. Potential table in it is not dropped.");
```
rxin: This will log for every drop table command when tachyon is not enabled. Can we make it less invasive? I am not even sure if we need this ...

src/tachyon_enabled/scala/shark/tachyon/TachyonUtilImpl.scala
```diff
@@ -35,12 +35,22 @@ class TachyonUtilImpl(val master: String, val warehousePath: String) extends Tac
...
   def getPath(tableName: String): String = warehousePath + "/" + tableName
 
+  override def tachyonEnabled(): Boolean = true
+
   override def tableExists(tableName: String): Boolean = {
     client.exist(getPath(tableName))
   }
 
+  override def dropTable(tableName: String): Boolean = {
+    return client.delete(getPath(tableName), true)
```
rxin: In what conditions would this return false? 

And if it is false, should we do any error handling in unpersist?

haoyuan: It should be true in normal condition. However, in following conditions, it will return false
1. two clients are dropping a table at the same time, one client will fail. 
2. unknown reason, it is like failed to remove a file from a file system.

Will add some handling in SharkEnv.

rxin: thanks. 

can you add the parameter name to the "true" and remove the "return"

```
client.delete(getPath(tableName), whateverthenameis=true)
```

and add the above conditions as inline comments?

src/tachyon_enabled/scala/shark/tachyon/TachyonTableRDD.scala
```diff
@@ -63,19 +61,21 @@ class TachyonTableRDD(path: String, @transient sc: SparkContext)
     val rawTable: RawTable = SharkEnvSlave.tachyonUtil.client.getRawTable(path)
     val buffers = Array.tabulate[ByteBuffer](rawTable.getColumns()) { columnIndex =>
       val fp = rawTable.getRawColumn(columnIndex).getPartition(theSplit.index, true)
-      var buf: ByteBuffer = fp.readByteBuffer()
+      var buf = fp.readByteBuffer()
       if (buf == null && fp.recache()) {
```
rxin, line 66: Can you add some comment on why this extra attempt?

src/tachyon_enabled/scala/shark/tachyon/TachyonTableRDD.scala
```diff
@@ -63,19 +61,21 @@ class TachyonTableRDD(path: String, @transient sc: SparkContext)
...
         buf = fp.readByteBuffer()
       }
       if (buf == null) {
         // TODO Log this. Reading data from remote is bad.
-        buf = ByteBuffer.allocate(fp.length().toInt)
+        var data = ByteBuffer.allocate(fp.length().toInt)
```
rxin, line 73: Change the var to val, and add logging (remove the TODO).

src/tachyon_enabled/scala/shark/tachyon/TachyonTableRDD.scala
```diff
@@ -63,19 +61,21 @@ class TachyonTableRDD(path: String, @transient sc: SparkContext)
...
         val is = fp.getInStream(ReadType.CACHE)
-        is.read(buf.array)
+        is.read(data.array)
         is.close()
-        buf.limit(fp.length().toInt)
+        data.limit(fp.length().toInt)
```
rxin, line 77: Just curious in what case do we need the limit here? The limit by default is fp.length, no?

src/main/scala/shark/SharkEnv.scala
```diff
@@ -110,6 +110,15 @@ object SharkEnv extends LogHelper {
   val addedJars = HashSet[String]()
 
   def unpersist(key: String): Option[RDD[_]] = {
+    if (SharkEnv.tachyonUtil.tachyonEnabled() && SharkEnv.tachyonUtil.tableExists(key)) {
+      logInfo("Table " + key + " is in Tachyon.");
```
rxin: remove this logInfo

project/SharkBuild.scala
```diff
@@ -97,6 +99,6 @@ object SharkBuild extends Build {
       "org.scalatest" %% "scalatest" % "1.9.1" % "test",
       "junit" % "junit" % "4.10" % "test",
       "com.novocode" % "junit-interface" % "0.8" % "test") ++
-      (if (TACHYON_ENABLED) Some("org.tachyonproject" % "tachyon" % "0.3.0-SNAPSHOT") else None).toSeq
+      (if (TACHYON_ENABLED) Some("org.tachyonproject" % "tachyon" % "0.3.0-SNAPSHOT" excludeAll(excludeKyro) ) else None).toSeq
```
rxin: Wouldn't this be a problem for Tachyon? Tachyon's kryo still uses de.javakaffee's library.

haoyuan: It's okay, since only server side uses it.

rxin:
Jenkins, test this please.

src/tachyon_enabled/scala/shark/tachyon/TachyonTableRDD.scala
haoyuan, line 77: in all cases. 

since we use data.array to put data into the underlayer array without modifying data.limit, we need to set it manually.

src/main/scala/shark/SharkEnv.scala
haoyuan: yes. will remove this later with other stuff in a commit.

project/SharkBuild.scala
```diff
@@ -32,13 +32,15 @@ object SharkBuild extends Build {
   val HADOOP_VERSION = "1.0.4"
 
   // Whether to build Shark with Tachyon jar.
-  val TACHYON_ENABLED = false
+  val TACHYON_ENABLED = true
```
rxin: Let's make it false by default for now (until there is a stable version of Tachyon in Jenkins). Otherwise, we won't be able to run the CI scripts anymore.

haoyuan:
Jenkins, test this please.

haoyuan:
Jenkins, test this please.

haoyuan:
Jenkins, test this please.

PR #111, opened 2013-07-30
Shark Scala API update

rxin:
@mateiz, @jhartlaub, @markhamstra

Please take a look at the new API change. There are multiple changes:
1. TableRDD has a new field tableDesc, which includes some schema definition. Previously it included scheme description that uses some Hive internal data structure. This is now gone and replaced with Shark's own data structure to represent the types.
2. Added a executeSql command to SharkContext, which returns a tuple (Schema, ResultData).
   Question 1: Should I return a ResultSet class with a scheme field and a data field rather than using Scala's tuple 2?
   Question 2: Should executeSql replace the existing sql() call?
3. Added a bunch of methods in Row to deal with complex data types. The problem with those is they are exposing some internal Hive data structures, and those can change and are not very well documented. Should I just disable those for now until we can properly generate all data structures in Shark using code gen?

In the long run, I would like to make the Row class serializable and have a proper equals / hashcode implementation, so we can run all Spark transformations/actions on those (collect, reduce, etc).  But we won't be able to do that until we do full code generation for all operator data structures in Shark.


src/test/scala/shark/SQLSuite.scala
```diff
@@ -79,7 +79,9 @@ class SQLSuite extends FunSuite with BeforeAndAfterAll {
   }
 
   private def expectSql(sql: String, expectedResults: Array[String], sort: Boolean = true) {
-    val results = if (sort) sc.sql(sql).sortWith(_ < _) else sc.sql(sql)
+    val sharkResults: Array[String] = sc.executeSql(sql)._2.map(_.mkString("\t")).toArray
+    val results = if (sort) sharkResults.sortWith(_ < _) else sharkResults
```
markhamstra, line 83: The extra boilerplate to extract the result array is kind of unfortunate.  It would be better if there were sufficient implicits available to make possible something like `val (schema, results) = sc.executeSql(sql)` (or `val (_, results) = ...` in this case.)

src/main/scala/shark/SharkContext.scala
```diff
@@ -104,6 +103,29 @@ class SharkContext(
     }
   }
 
+  def executeSql(cmd: String, maxRows: Int = 1000): (Array[ColumnDesc], Seq[Array[Object]]) = {
```
mateiz: I think this should return a special class to represent the result set. The reason is that a tuple will be harder to work with in Java, and will also be harder to extend if we want to return more information over time.

Also, the doc comment for this method should say that it's going to collect the data and return it locally (as opposed to sql2rdd). It may also be good to just call it "runSql" or "execute" if those aren't taken.

src/main/scala/shark/api/DataTypes.java
```diff
@@ -0,0 +1,104 @@
...
+  };
+
+  private static Map<String, DataType> hiveTypes = new HashMap<String, DataType>();
+  static {
+    for (DataType type: types) {
+      hiveTypes.put(type.hiveName, type);
+    }
+  }
+
+  static class DataType implements Serializable {
```
mateiz: Any reason why this is a nested class? Seems cleaner to make it top-level, or maybe make the static things in DataTypes be part of DataType.

src/main/scala/shark/api/JavaSharkContext.scala
```diff
@@ -22,8 +22,8 @@ import java.util.{List => JList}
 
 import scala.collection.JavaConversions._
 
-import spark.api.java.JavaSparkContext
 import shark.SharkContext
+import spark.api.java.JavaSparkContext
 
 class JavaSharkContext(val sharkCtx: SharkContext) extends JavaSparkContext(sharkCtx) {
```
mateiz, line 28: It would be nice if this had an executeSql equivalent as well.

src/main/scala/shark/api/Row.scala
```diff
@@ -55,64 +68,64 @@ class Row(val rawdata: Any, val colname2indexMap: Map[String, Int], val oi: Stru
...
-  def getString(field: Int): String = {
+  def getMap(field: Int): java.util.Map[_, _] = {
     val ref = oi.getAllStructFieldRefs.get(field)
     val data = oi.getStructFieldData(rawdata, ref)
-    ref.getFieldObjectInspector.asInstanceOf[StringObjectInspector].getPrimitiveJavaObject(data)
+    ref.getFieldObjectInspector.asInstanceOf[MapObjectInspector].getMap(data)
   }
 
-  def getTimestamp(field: Int): java.sql.Timestamp = {
+  def getStruct(field: Int): java.util.List[Object] = {
```
mateiz, line 126: For this, do we really want to return a list? Won't it lose the information about names of the fields? I guess it can be okay if they see the struct's schema outside.

PR #113, opened 2013-08-03
Hive 0.11 compatibility

alupher:
Initial commit of support for Hive 0.11. This compiles and runs as expected, passing most regression tests, though more thorough testing is still in progress. Does not yet include support for new Hive 0.11 operators (e.g. PTFOperator) or new stats features.

### DIFF
--- a/src/main/scala/shark/parse/SharkExplainSemanticAnalyzer.scala
+++ b/src/main/scala/shark/parse/SharkExplainSemanticAnalyzer.scala
@@ -62,7 +62,7 @@ class SharkExplainSemanticAnalyzer(conf: HiveConf) extends ExplainSemanticAnalyz
     }
 
     val task = TaskFactory.get(
-      new SharkExplainWork(ctx.getResFile().toString(), tasks, childNode.toStringTree(), extended),
+      new SharkExplainWork(ctx.getResFile().toString(), tasks, childNode.toStringTree(), sem.getInputs(), extended),
       conf)
 
     rootTasks.add(task)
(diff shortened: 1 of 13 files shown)


src/main/scala/shark/parse/SharkExplainSemanticAnalyzer.scala
```diff
@@ -62,7 +62,7 @@ class SharkExplainSemanticAnalyzer(conf: HiveConf) extends ExplainSemanticAnalyz
     }
 
     val task = TaskFactory.get(
-      new SharkExplainWork(ctx.getResFile().toString(), tasks, childNode.toStringTree(), extended),
+      new SharkExplainWork(ctx.getResFile().toString(), tasks, childNode.toStringTree(), sem.getInputs(), extended),
```
rxin: wrap this at 100 col

src/test/java/shark/SharkQTestUtil.java
```diff
@@ -89,6 +93,7 @@ public void cliInit(String tname, boolean recreate) throws Exception {
 
   @Override
   public int executeClient(String tname) {
+    LOG.info("Begin query: " + tname);
```
rxin, line 96: Change this to "Begin test: "

alupher, line 96: Looks like Hive uses "Begin query" (see line 126 of https://github.com/apache/hive/blob/branch-0.11/ql/src/test/templates/TestCliDriver.vm) -- probably should keep the same for consistency.

rxin:
Merging this in so we can update the hive 0.11 branch. Thanks, Antonio!

PR #114, opened 2013-08-06
Console prints time taken for queries in remote mode.

lin-zhao:
If you run shark shell in remote mode, with -h and -p, the console doesn't print time taken. This is useful for users to evaluate query performance.


rxin:
Thanks, Lin!

PR #115, opened 2013-08-07
New bin/run-tests-from-scratch to be used by Jenkins or new users as reference

andyk:
Add a new bash script bin/run-tests-from-scratch that sets up a Shark environment (including the "master" branch of Spark) from scratch and runs the unit tests. The script requires JAVA_HOME and SCALA_HOME to be set but does all of the rest of the work for you, including:
1. Download and build Spark master
2. Download/untar/use whichever Hadoop is set to whatever by HADOOP_VERSION env var (or "0.20.205.0" is used if HADOOP_VERSION is unset)
3. Download patched Hive from github/amplab and build it
4. Run Hive tests with `ant test` until we see "[junit]" string and then kill those tests (kill all children tests too, which is a bit hairy)
5. Run Shark's `sbt/sbt test`
6. Run the Hive tests that Shark is known to pass via `TEST_FILE=src/test/tests_pass.txt ./bin/dev/test`

Flags can be passed to the tests to make life easier for developers by allowing them to skip portions of the setup that have completed successfully. Pass -h to see a list of available flags.

Once this is committed, Jenkins will use this script to run all of the tests every time a commit is made to master and I'll also set up a pull request builder that will run this script when asked to via github comments.

You can actually see it working in a limited way in Jenkins already at https://amplab.cs.berkeley.edu/jenkins/job/Shark-Master-Hadoop1


bin/run-tests-from-scratch
```diff
@@ -0,0 +1,252 @@
...
+EOF
+exit
+}
+
+SKIP_SPARK=false
+SKIP_HIVE=false
+SKIP_HADOOP=false
+SKIP_TESTS=false
+
+while getopts "hsed" opt; do
```
rxin: This is pretty fancy :)

src/test/scala/shark/SQLSuite.scala
```diff
@@ -139,6 +139,8 @@ class SQLSuite extends FunSuite with BeforeAndAfterAll {
   //////////////////////////////////////////////////////////////////////////////
   // sorting
   //////////////////////////////////////////////////////////////////////////////
+
+  /* Commenting this out since Linux file systems don't return results alphabetically, while OS X and HDFS does.
```
rxin: Let's use the standard scalatest way of disabling a test by changing "test" to "ignore".

http://doc.scalatest.org/1.9.1/index.html#org.scalatest.FunSuite

rxin:
Hi Andy,

This looks really good. Thanks a lot for doing this! 

In addition to the nitpick I made, can we also rename the test file from bin/run-tests-from-scratch to bin/dev/run-tests-from-scratch.sh?

bin/run-tests-from-scratch
```diff
@@ -0,0 +1,252 @@
...
+   -h      Show this message
+   -s      Skip Spark (downloading and building)
+   -e      Skip Hive (downloading and building)
+   -d      Skip Hadoop (downloading and buildling)
+   -t      Compile Shark but skip compiling and running Tests.
+EOF
+exit
+}
+
+SKIP_SPARK=false
```
rxin: Actually one other thing.  I am just thinking about how we can make it easy for people that are folking Shark to use this script. They might have their own git repositories for Spark and Hive they have to define. And we want to make it easy to merge stuff so the configuration can stay in a separate file.

Can I suggest this small change?

First, source the run-tests-from-scratch-config.sh file for private configurations if the file exists. 

Then, make the Spark git repo + spark branch and Hive git repo + branch into their own variables, and set default values if they are not set already. 

This would make this script very portable, and configurations can be done entirely in the private run-tests-from-scratch-config.sh file.

src/test/scala/shark/SQLSuite.scala
andyk: Done.

andyk:
Hey @rxin, regarding whether to have a .sh on it or not. One popular convention is to have executable files (e.g., files in bin) not have a suffix and non-executable files (e.g., files in conf) have a suffix. We've been followed this convention in Shark so far (i.e. bin/shark, bin/shark-shell, etc.), any reason you want to deviate in this case?

rxin:
Actually that makes sense. Let's just move that file to bin/dev

bin/run-tests-from-scratch
andyk: I've been thinking about this and I'm not sure I believe adding a config file will make life easier for orgs who want to set up Jenkins for their forks of Shark. Maybe some other users can chime in here.

My thinking is that they would need to modify and commit this file to their fork of the repo so that Jenkins would check it out every time it does a build, and this is different than the pattern we have for the [spark|shark]-env.sh config files and I think that would make using a config file a bit more confusing.

I do definitely agree with you that it should be more obvious that orgs can configure things to work in their private environments via env variables. We can recommend either (a) just export the appropriate env vars in the Jenkins project before calling bin/run-tests-from-scratch and or (b) create and commit (to their private repo or to somewhere they can fetch via wget) a text file containing these and source that file before calling bin/run-tests-from-scratch.

So, based on those thoughts, I think we should:

1) Make sure all variables that orgs might want to use custom values for respect having their default values being overridden when env vars are set. E.g. the following would have to be changed: HIVE_WAREHOUSE, HIVE_DEV_HOME, HIVE_HOME, SPARK_MEM, SPARK_MASTER_MEM, ...

2) Clearly document these variables. This includes (1) in the comment at the top of the script, (2) in the usage text that prints with -h, and (3) by moving them all to the top of the file (and having parsing and checking of them come in a section below where they are all set instead of inline with them as it is now).

What do you think?

andyk:
Ok @rxin, I think I've addressed all of your feedback and made this much easier for developers to pick up and use when they want to get their Shark env setup quickly or set up their own CI for Shark.

Let me know if think it needs any other refactoring or I missed any of your comments. Else, I think it's ready to merge.

rxin:
Thanks, Andy! Merging this now...

PR #116, opened 2013-08-07
Refactor Column Building / Iteration in preparation for compression

harsha2010:


src/main/scala/shark/memstore2/column/ColumnBuilders.scala
```diff
@@ -0,0 +1,207 @@
...
+  
+  override def append(v: Timestamp) = {
+    buffer.putLong(v.getTime())
+    buffer.putInt(v.getNanos())
+  }
+}
+
+class BinaryColumnBuilder extends NullableColumnBuilder[BytesWritable] {
+
+  _stats = null
```
rxin, line 166: Let's create a noop stat class so we can avoid using nulls here. Also avoid checking for nulls in append.

src/main/scala/shark/memstore2/column/ColumnIterators.scala
```diff
@@ -0,0 +1,189 @@
...
+    val _currentLen = buffer.getInt
+    if (_currentLen >= 0) {
+      val newBytes = new Array[Byte](_currentLen)
+      buffer.get(newBytes, 0, _currentLen)
+      bytesFld.set(_writable, newBytes)
+      lengthFld.set(_writable, _currentLen)
+    }
+  }
+
+  override def current = if (_currentLen >= 0) _writable else null
```
rxin, line 135: This shouldn't be needed anymore since null is handled outside now, no?

src/main/scala/shark/memstore2/column/ColumnIterators.scala
```diff
@@ -0,0 +1,189 @@
...
+}
+
+class StringColumnIterator(val buffer: ByteBuffer) extends ColumnIterator {
+  private val _writable = new Text
+  private var _currentLen = 0
+  private val bytesFld = {
+    val f = _writable.getClass().getDeclaredField("bytes")
```
rxin, line 142: Actually if you are doing this, why don't you just write directly to the bytes array instead of creating a new byte array?

harsha2010, line 142: good point :), i started out to do that and forgot!

src/main/scala/shark/memstore2/column/ColumnIterators.scala
harsha2010, line 135: yes

src/main/scala/shark/memstore2/column/ColumnBuilders.scala
harsha2010, line 166: yes, makes sense

src/main/scala/shark/memstore2/column/ColumnIterators.scala
rxin, line 142: make sure you check the length. u might have to reallocate a new array if the length is too short

PR #118, opened 2013-08-08
Automatically download Scala in CI.

rxin:


bin/dev/run-tests-from-scratch
```diff
@@ -8,6 +8,12 @@
 # Run this file with -h (or see the usage() function below) for details about
 # configuration options.
 
+if [ "x$SHARK_PROJ_DIR" == "x" ] ; then
```
andyk: Moving this up makes complete sense. Now that I think of it, it seems like this would fit in even more naturally between where we set the default variables and where we check/set up JAVA_HOME, and then we can pull out "`pwd`/../.." and set it up as SHARK_PROJ_DIR_DEFAULT and it to the usage string where we print the default values we use. I should probably have done this from the beginning, since it's a (needlessly) secret configuration option as we have it right now. Mind doing this?

bin/dev/run-tests-from-scratch
```diff
@@ -219,6 +218,16 @@ if [ -e "${SHARK_PROJ_DIR}/conf/shark-env.sh" ] ; then
 fi
 
 ####################################################################
+# Download Scala if SCALA_HOME is not specified.
+####################################################################
+if [ "x$SCALA_HOME" == "x" ] ; then
+  wget $SCALA_DOWNLOAD_PATH
```
andyk, line 229: It just occurred to me that we should probably do a check of the return status of wget with something like `if (($? != 0)); then echo Err msg; exit -1 ; fi` or possibly make sure scala*tgz exists after the wget command. This is more defensive than we were being in the current use of wget for hadoop (but you could update that spot too maybe?)

rxin, line 229: Wouldn't the program just quit if wget fails?

bin/dev/run-tests-from-scratch
```diff
@@ -8,18 +8,37 @@
...
 #####################################################################
-SPARK_MEM_DEFAULT=8g
-SPARK_MASTER_MEM_DEFAULT=8g
+SPARK_MEM_DEFAULT=4g
+SPARK_MASTER_MEM_DEFAULT=4g
 SPARK_GIT_URL_DEFAULT="https://github.com/mesos/spark.git"
 HIVE_GIT_URL_DEFAULT="https://github.com/amplab/hive.git -b shark-0.9"
 HADOOP_VERSION_DEFAULT="0.20.205.0"
 HADOOP_MAJOR_VERSION_DEFAULT=1
 SPARK_KV_JAVA_OPTS_DEFAULT=("-Dspark.local.dir=/tmp " "-Dspark.kryoserializer.buffer.mb=10 ")
```
andyk: Can you add a `SHARK_PROJ_DIR_DEFAULT=`pwd`/../..` here? then update line 27 to use it below and update the text in the usage() function under the "Optional configuration environment variables:" heading to include it too?

bin/dev/run-tests-from-scratch
```diff
@@ -104,9 +113,9 @@ source it before you run this script in your CI project.
 
 Required Options:
   JAVA_HOME (Required)
-  SCALA_HOME (Required)
 
 Optional configuration environment variables:
+  SCALA_HOME
```
andyk: How about appending `(default: Scala version ${SCALA_VERSION} will be downloaded and used)` to this line?

bin/dev/run-tests-from-scratch
andyk: Actually, it would be better to make sure these all come in the same order that they come in the next (setup section) as well as the printout in the usage() message.

bin/dev/run-tests-from-scratch
andyk, line 229: hmm, yeah, you're right. It seemed a little dirty at first to use that -e flag and i thought we might want to stop using it at some point, but after thinking about it a little more, I think it actually might be the right thing to use in this script to avoid having ugly checks of $? all the time (like what I was suggesting here) since we really do want the fail fast behavior throughout most of this type of script.

andyk:
I think it looks good after you add that default value and clean up the ordering of the default values to be the same in all three of the sections they show up in, i.e.: (1) where initialized; (2) where used; and (3) where printed in usage().

PR #121, opened 2013-08-12
Bugfix: "-t" flag was getting ignored. Also, "sbt/sbt compile" -> "sbt/sbt publish-local"

andyk:


rxin:
Hi Andy,

Why did you change "compile" to "publish-local"?

andyk:
I thought it might be the more usual use-case (at least that was my use case last night). I'm fine sticking with compile if you think that's better.

PR #122, opened 2013-08-13
Refactor Column Builders, Iterators., Add Compression API for Shark, with RLE implementation

harsha2010:
This pull request refactors the column builder and iterator APIs to enable pluggable compression. A compression API is introduced with a RLE implementation.


rxin:
Jenkins, test this please.

rxin:
Jenkins, test this please.

rxin:
This is merged manually now

mbbui:
I wonder why EWAHCompressedBitmap is no longer used to encode null position now?

rxin:
For simplicity for now. We want to add some sort of null compression back eventually, but I don't think we need a full-fledged ewah. We also want the bitmap to be more bytebuffer friendly (so we don't need a full deserialization step).

Let me know if you are interested in working on it.

mbbui:
Yes, I am interested in helping with this. It does not seem to be too complicated, do you think so?

rxin:
Yea - it should be very isolated (changing a small number of files). It would require understanding the algorithm ewah uses (which should also be relatively simple) and making that work directly against bytebuffers.

PR #126, opened 2013-08-16
Reorder materialization and cache insertion

harsha2010:
In case of failures during materialization the current ordering means we cannot clean up RDDs that are cached when the table creation fails.
This patch fixes the cleanup issue by ordering table creation first followed by RDD materialization.


rxin:
I merged this manually because there was a small conflict.

PR #127, opened 2013-08-18
Column Pruning

harsha2010:
This patch prunes away columns that are not used in a query so that the  table partition iterator can skip over those columns while iterating


rxin:
Jenkins, test this please.

Jenkins, add to whitelist.

rxin:
Jenkins, test this please.

rxin:
Thanks!

PR #129, opened 2013-08-18
Refactor column stats., add between and in pruning. introduce bloom filter and murmur hash implementation 

harsha2010:
1. Refactors the column stats API to more easily apply pruning. 
2. Between and In clause are now pruned correctly
3. Introduce Bloom Filter implementation with a MurmurHash3 implementation so subsequently we can start pruning of high cardinality columns. (The Bloom filter is not plugged into column stats yet, since column stats is collected only on the master). A subsequent patch will introduce either a) the ability to specify turning on/ off bloom filter usage per column, or b) move portion of the stats to the RDDs themselves after which we can start using such filters for more sophisticated pruning


rxin:
Jenkins, test this please.

rxin:
Thanks!

PR #134, opened 2013-08-20
Fix SHARK-175: TS supports object inspectors initialized on partition metadata

alupher:
This allows the previously failing 'partition_vs_table_metadata' test to pass. This problem was due to object inspectors in Hive 0.11 being initialized based on partition metadata instead of table metadata (see https://issues.apache.org/jira/browse/HIVE-3833).

Unit tests were run and no regression noted, but we may need to evaluate the performance impact of the extra serialization/deserialization steps on lines 291-293 (?)


alupher:
Merging pull request as per conversation with Reynold.

PR #135, opened 2013-08-22
Consistent naming for storage levels

amirshim:
Is there any reason for not making the string names consistent?


rxin:
Thanks - this looks good. In the long run we need to evaluate whether it makes sense to pass this storage level configuration this way.

PR #136, opened 2013-08-26
RDDTable (Convert RDD to Shark tables that can be queried in SQL)

rxin:


rxin:
Thanks for the comment. While I agree that it would be nice to support arbitrary types, I am also concerned about exposing the internals of TablePartitionBuilder to the user. I am not sure what the best way to do this is right now.

mateiz:
Maybe one compromise would be to add a version that works on Seq? Then people doing custom serialization can return a Seq instead of a tuple, which might be more convenient and would allow you to have more than 22 columns. It shouldn't be too hard to map() your custom objects to either a Seq or a tuple.

aht:
Very nice feature!

One case that we'd like to use a lot is to convert some RDD[Array[Object]] to a table so having a version that work on Seq would be great.

AndreSchumacher:
Hi all! What's the status of this PR? It would be a really nice feature. @rxin , are you still working on this?

rxin:
We are! Plan to do some updates and merge it for 0.8.1 release.

harveyfeng:
Manually merged this branch. The API's still undergoing a few changes - PRs should be against the master from now on.

PR #140, opened 2013-08-28
Throw exception if Hive thinks it's running in local mode

pwendell:
This throws a descriptive error message if a user tries to run Shark in a way that will result in query responses being silently dropped.

### DIFF
--- a/src/main/scala/shark/execution/FileSinkOperator.scala
+++ b/src/main/scala/shark/execution/FileSinkOperator.scala
@@ -120,11 +120,9 @@ class FileSinkOperator extends TerminalOperator with Serializable {
     val inputRdd = if (parentOperators.size == 1) executeParents().head._2 else null
     val rdd = preprocessRdd(inputRdd)
 
-    val master = rdd.context.master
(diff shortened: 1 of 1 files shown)


src/main/scala/shark/execution/FileSinkOperator.scala
```diff
@@ -119,6 +120,18 @@ class FileSinkOperator extends TerminalOperator with Serializable {
     val inputRdd = if (parentOperators.size == 1) executeParents().head._2 else null
     val rdd = preprocessRdd(inputRdd)
 
+    val master = rdd.context.master
+    val sparkIsLocal = (master == "local" || master.startsWith("local["))
```
rxin: Also allow local-cluster[

rxin:
To make this more robust, Patrick is going to expose the isLocal attribute in SparkContext and update this pull request.

pwendell:
Jenkins, test this please.

pwendell:
Hey @rxin - this should be good to go now. Mind taking a look and/or merging?

rxin:
Thanks, Patrick!

PR #141, opened 2013-08-28
Set the RDD name in Spark when creating an RDD-backed table

pwendell:
Tested this locally and it worked when creating a cached table.


rxin:
Thanks, Patrick!

PR #144, opened 2013-09-01
Try to fix configuration problem for user

pwendell:


pwendell:
Jenkins, test this please.

colorant:
@pwendell just curious, on what condition this might happen? say shim detect local while rdd not run local? I run into this warning when try to run shark on Yarn 2.2. 

the task runs with error : cluster.ClusterTaskSetManager: Loss was due to java.lang.IllegalArgumentException: Wrong FS: file:/tmp/root/hive_2013-12-11_11-17-43_295_666389238294242717/_tmp.-ext-10001, expected: hdfs://public-cluster [duplicate 3]

and there are warning before it.

13/12/11 11:17:45 WARN execution.FileSinkOperator: Hive Hadoop shims detected local mode, but Shark is not running locally.
13/12/11 11:17:45 WARN execution.FileSinkOperator: Setting mapred.job.tracker to 'Spark_1386731865096' (was 'local')

Not sure what lead to this issue. btw. I do have hadoop conf in path, and hdfs works. I don't run Hive directly, since hive 0.9 won't work on yarn 2.2 I think. but since spark now run on yarn 2.2. I think shark should be ok.

PR #145, opened 2013-09-03
Some misc. comments in TableScanOperator and MemoryMetaDataManager

harveyfeng:
Added while looking into support for Hive-partitioned cache tables...


rxin:
Thanks!

PR #150, opened 2013-09-06
Column compression cleanup

rxin:
- In dictionary encoding, the dictionary itself doesn't need to store the index (the values are stored in order of the index)
- In dictionary encoding decoder, uses an array to replace the original hash table so dictionary lookup is always O(1)
- Remove the position argument for extract/extractInfo methods in ColumnType
- init() is called in a ColumnIterator's constructor, and removed computeNext()
- Corrected various tests. Some tests were testing for the wrong thing (e.g. a few tests used null count for compression type ...)
- Various other minor changes to improve code readability and test readability

### DIFF
--- a/src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
+++ b/src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
@@ -13,7 +13,13 @@ import shark.memstore2.column.Implicits._
 class CompressedColumnIteratorSuite extends FunSuite {
 
   /**
-   * Generic tester across types and encodings
+   * Generic tester across types and encodings. The function applies the given compression
+   * algorithm on the given sequence of values, and test whether the resulting iterator gives
+   * the same sequence of values.
+   *
+   * If we expect the compression algorithm to not compress the data, we should set the
+   * shouldNotCompress flag to true. This way, it doesn't actually create a compressed buffer,
+   * but simply tests the compression ratio returned by the algorithm is >= 1.0.
    */
   def testList[T, W](
       l: Seq[T],
--- a/src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
+++ b/src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
@@ -1,3 +1,20 @@
+/*
+ * Copyright (C) 2012 The Regents of The University California.
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package shark.memstore2.column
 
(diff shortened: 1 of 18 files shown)


src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
```diff
@@ -12,117 +12,116 @@ import shark.memstore2.column.Implicits._
...
+      t.append(item, b)
+      algo.gatherStatsForCompressibility(item, t.asInstanceOf[ColumnType[Any, _]])
     }
     b.limit(b.position())
     b.rewind()
-    val compressedBuffer = algo.compress(b, u)
+
     if (shouldNotCompress) {
       assert(algo.compressionRatio >= 1.0)
-      info("CompressionRatio " + algo.compressionRatio)
```
junkumar, line 44: Why remove this information from the test output?

src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
```diff
@@ -12,117 +12,116 @@ import shark.memstore2.column.Implicits._
...
+        case _       => throw new UnsupportedOperationException("Unsupported compression type " + t)
       }
 
       l.foreach { x =>
         iter.next()
-        assert(compareFunc(u.get(iter.current, oi), x))
-        // assert(u.get(iter.current, oi) === x)
+        assert(compareFunc(t.get(iter.current, oi), x))
       }
-      assert(false === iter.hasNext) // no extras at the end
```
junkumar, line 63: I think the triple = gives a better message upon failure. Just checking the function's return dumps a harder to read stack trace on the console.

src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
```diff
@@ -12,117 +12,116 @@ import shark.memstore2.column.Implicits._
...
+      shouldNotCompress = true)
   }
 
   test("RLE - should not compress") {
-    val alternating = Range(0, Short.MaxValue-1, 1).flatMap { s => List(1, s) }
-    val hugeList = List.concat(iList, alternating, iList)
-    assert(hugeList.size === (8 + 2*(Short.MaxValue-1)))
-    testList(hugeList, INT, new RLE(), (a: Int, b: Int) => a == b, true)
+    val ints = Range(0, Short.MaxValue.toInt + 1)
+    testList(ints, INT, new RLE, (a: Int, b: Int) => a == b, shouldNotCompress = true)
```
junkumar: How about keeping the (shouldNotCompress = true) test at (Short.MaxValue.toInt + 1) as well as the 
 (shouldNotCompress = false) test at (Short.MaxValue.toInt - 1)

This way we test the limit from both directions.

src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
rxin, line 63: It tells you a message like "true does not equal false". I don't think that is more helpful than the assertion is not true ...

src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
rxin: Good idea. (Actually I still have the (Short.MaxValue.toInt - 1) test. Look about 15 lines up)

src/test/scala/shark/memstore2/column/CompressedColumnIteratorSuite.scala
rxin, line 44: Actually good idea - I am adding this back

PR #155, opened 2013-09-11
Some refactoring and initial code for Hive-partitioned tables.

harveyfeng:
There are a few TODOs in the comments to address sketchy additions/API changes and possible refactorings.
The most notable one is SharkSemanticAnalyzer#analyzeCreateTable() regarding how Shark should create a cached table without CTAS.

### DIFF
--- a/project/SharkBuild.scala
+++ b/project/SharkBuild.scala
@@ -45,6 +45,9 @@ object SharkBuild extends Build {
   val excludeKyro = ExclusionRule(organization = "de.javakaffee")
   val excludeHadoop = ExclusionRule(organization = "org.apache.hadoop")
   val excludeNetty = ExclusionRule(organization = "org.jboss.netty")
+  val excludeJackson = ExclusionRule(organization = "org.codehaus.jackson")
+  val excludeAsm = ExclusionRule(organization = "asm")
+  val excludeSnappy = ExclusionRule(organization = "org.xerial.snappy")
 
   def coreSettings = Defaults.defaultSettings ++ Seq(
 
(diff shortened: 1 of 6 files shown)


src/main/scala/shark/memstore2/MemoryMetadataManager.scala
```diff
@@ -27,22 +27,46 @@ import org.apache.spark.storage.StorageLevel
 
 import shark.SharkConfVars
 
-
+// TODO(harvey): Re-evaluate the interfaces to this class. For example, add() could be renamed to
+//               addCreatedTable().
```
harveyfeng: Are there plans to add non-tables in MemoryMetadataManager? It was (previously) pretty general.

src/main/scala/shark/parse/SharkSemanticAnalyzer.scala
```diff
@@ -354,6 +328,91 @@ class SharkSemanticAnalyzer(conf: HiveConf) extends SemanticAnalyzer(conf) with
...
+        }
+
+        // In Hive, a CREATE TABLE command is handled by a DDLTask, which in this case, is created
+        // by the Hive SemanticAnalyzer's genMapRedTasks and not Hive's DDLSemanticAnalyzer. Since
+        // creating tables in Shark doesn't involve too much overhead (we don't support features
+        // such as indexing), just directly update the Shark MemoryMetaDataManager in this method.
+
+        // Make sure that the table exists.
+        // TODO(harvey): This might have to go in a SparkDDLTask wrapper.
+        // SharkEnv.memoryMetadataManager.add(tableName, isHivePartitioned)
```
harveyfeng: This line can't be called from SharkSemanticAnalyzer. The Hive DDLTask created to handle CREATE TABLE could throw an exception, which can only be caught in the SharkDriver after Driver#execute(), so there's no knowing at the semantic analysis phase whether the table-to-be-created will actually be created.

harveyfeng:
Sorry about that, tests should all pass now....

src/main/scala/shark/memstore2/MemoryMetadataManager.scala
rxin: Yea I don't think so. We can specialize it a little bit more. 

Maybe we can force the users to specify the database name, the table name.

src/main/scala/shark/parse/SharkSemanticAnalyzer.scala
harveyfeng: (Hacky)
Another alternative is to have an interceptor closure that's a HiveDriverRunHook.
The interceptor closure will only be used for a CREATE TABLE.
It can be passed through HiveConf.ConfVars.HIVE_DRIVER_RUN_HOOKS in SharkDriver when it detects a CREATE TABLE. The 'interceptor' refers to how the created closure could wrap any Hook the user defines.

Driver#run() gets run hooks using this:

```
  driverRunHooks = getHooks(HiveConf.ConfVars.HIVE_DRIVER_RUN_HOOKS, HiveDriverRunHook.class);
```

PR #156, opened 2013-09-11
Fix name for non assembly jar.

colorant:
Fix the name for non-assembly jar. Tweak of jarName is adequate for the fat jar.


rxin:
Jenkins, test this please.

rxin:
Thanks. Jenkins is not responding so I tested this manually.

PR #160, opened 2013-09-13
Exclude asm, jackson, and netty from hadoop client. Changed jets3t's version to 0.7.1.

rxin:
Need to apply this to branch-0.8 as well.


rxin:
Jenkins, test this please.

PR #161, opened 2013-09-13
Fix bug of join filter doesn't work in JoinOperator

chenghao-intel:
JoinFilters were ignored in JoinOperator, queries like:

"SELECT src1.key as k1, src1.value as v1, src2.key as k2, src2.value as v2 FROM src src1 RIGHT OUTER JOIN src src2 ON (src1.key = src2.key and src2.key < 20 and src1.key < 20);"

The filter conditions didn't take effect. This patch will solve the bug. 

BTW: I notice that the Hive-0.9 seems give the wrong result for above query.


rxin:
Jenkins, test this please.

rxin:
Thanks for the fix. This brings 3 regression tests failures. Can you take a look?

There were 3 failures:
1) testCliDriver_auto_join20(org.apache.hadoop.hive.cli.TestCliDriver)junit.framework.AssertionFailedError: Unexpected exception
See build/ql/tmp/hive.log, or try "ant test ... -Dtest.silent=false" to get more logs.
    at org.apache.hadoop.hive.cli.TestCliDriver.testCliDriver_auto_join20(TestCliDriver.java:3139)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
2) testCliDriver_join20(org.apache.hadoop.hive.cli.TestCliDriver)junit.framework.AssertionFailedError: Unexpected exception
See build/ql/tmp/hive.log, or try "ant test ... -Dtest.silent=false" to get more logs.
    at org.apache.hadoop.hive.cli.TestCliDriver.testCliDriver_join20(TestCliDriver.java:20836)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
3) testCliDriver_join40(org.apache.hadoop.hive.cli.TestCliDriver)junit.framework.AssertionFailedError: Unexpected exception
See build/ql/tmp/hive.log, or try "ant test ... -Dtest.silent=false" to get more logs.
    at org.apache.hadoop.hive.cli.TestCliDriver.testCliDriver_join40(TestCliDriver.java:21958)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

You can run them with

```
TEST=auto_join20 ./bin/dev/test
```

chenghao-intel:
Sorry, for the delay.

Yeah, I am working on this. But join filter is a little bit more complicated than I thought, and also, we need to think about the SEMI join. Hope I can figure out how to fix that soon from reading the hive source code.

apivovarov:
it's recommended to replace "for" loops with "while" loops in Scala for speed.
Cheng, could you change it in your new code?

ss77892:
This bug looks similar to 
https://issues.apache.org/jira/browse/HIVE-1534
and that was fixed in hive 0.10 
Could you please check it with the shark's hive0.11 branch. On my box those queries works just fine on shark + hive0.11 without any modification on shark side.

apivovarov:
HIVE1534 says that it was fixed in 0.7.0
I just checked hive-0.11 - issue is still here

apivovarov:
I found it in JIRA  SHARK-69   https://spark-project.atlassian.net/browse/SHARK-69

As Reynold mentioned in Shark-69 -  push the filter to WHERE (which is actually much better for performance).

I think correct way to deal with it is to make query fail on sql validation step
So, if there is a filter predicate in "join on" then query should fail with error - filter conditions should go to "where" not to "join on"

apivovarov:
Cheng, what filter conditions didn't take effect?   src2 or src1?
if implementation is correct then
- right table src2.key < 20  should be ignored
- left table src1.key < 20 should take effect.

I checked shark-0.8 - it works exactly that
right table condition ignored tt2.c1 <= 1:
shark> select \* from tt1 right outer join tt2 on (tt1.c1 = tt2.c1 and tt2.c1 <= 1);
2   2
1   1
NULL    9
NULL    8

left table condition works tt1.c1 <= 1:
shark> select \* from tt1 right outer join tt2 on (tt1.c1 = tt2.c1 and tt1.c1 <= 1);
NULL    2
1   1
NULL    9
NULL    8

Nothing to fix here.

shark> select \* from tt1;  
1
2
3
4

shark> select \* from tt2;
1
2
8
9

chenghao-intel:
apivovarov, I did your queries on mysql, but got different result for the last query of below:

mysql> select \* from tt1;
+------+
| c1   |
+------+
|    1 |
|    2 |
|    3 |
|    4 |
+------+
4 rows in set (0.00 sec)

mysql> select \* from tt2;
+------+
| c1   |
+------+
|    1 |
|    2 |
|    8 |
|    9 |
+------+
4 rows in set (0.00 sec)

// same output with shark, which pushes down the filter "tt1.c1<=1" (tt1 is not on the right).
mysql> select \* from tt1 right outer join tt2 on (tt1.c1 = tt2.c1 and tt1.c1 <= 1);
+------+------+
| c1   | c1   |
+------+------+
|    1 |    1 |
| NULL |    2 |
| NULL |    8 |
| NULL |    9 |
+------+------+
4 rows in set (0.00 sec)

// different output with shark, as the filter "tt2.c1<=1" is the join filter (tt2 is on the right), and cannot be pushed down, and the filter expression is just ignored in shark code(in JoinOperator), which is a bug of shark.
mysql> select \* from tt1 right outer join tt2 on (tt1.c1 = tt2.c1 and tt2.c1 <= 1);
+------+------+
| c1   | c1   |
+------+------+
|    1 |    1 |
| NULL |    2 |
| NULL |    8 |
| NULL |    9 |
+------+------+
4 rows in set (0.00 sec)

From the code of JoinOperator.scala, I guess it also has bug in semi join. But I haven't figure out how to fix it yet.

apivovarov:
Interesting, I think it's just MySQL optimizer side effect - it adds additional condition "tt1.c1 <=1" to "join on"

Is there any formal description on how sql should work in this particular case?
without that we can not say if this is bug in shark or not. It might be implementation specific.

BTW postgres gives the same as MySQL.
postgres=# select \* from tt1 right outer join tt2 on (tt1.c1 = tt2.c1 and tt2.c1 <= 1);
  1 |  1
    |  2
    |  8
    |  9

ss77892:
Yeah, this is the shark bug.  How it should work. join takes all combinations between right table and left table:
    1 |    1 |
|    2 |    1 |
|    3 |    1 |
|    4 |    1 |
|    1 |    2 |
|    2 |    2 |
|    3 |    2 |
|    4 |    2 |
|    1 |    8 |
|    2 |    8 |
|    3 |    8 |
|    4 |    8 |
|    1 |    9 |
|    2 |    9 |
|    3 |    9 |
|    4 |    9 |
after that it applies the 'on' condition, so 
1) it should match tt1.c1=tt2.c1, so the list will be reduced to:
|    1 |    1 |
|    2 |    2 |
2) tt2.c2 <=1 will be applied:
|    1 |    1 |
3) since we have the outer join, we will add the missing elements from right table with nulls in other fields
|    1 |    1 |
| NULL |    2 |
| NULL |    8 |
| NULL |    9 |

Actually 1) and 2) checking at the same time, but I separated them for the visibility.

chenghao-intel:
Yes, that outer join filter (on filter) is quite confusing, http://www.daveturpin.com/2010/10/equality-filters-on-vs-where/ may gives some explanation.

BTW, even Hive gives the same output as mysql for "select \* from tt1 right outer join tt2 on (tt1.c1 = tt2.c1 and tt2.c1 <= 1);"

rxin:
Let's postpone this after 0.8, since I don't think it is a new bug, and we have actually documented it. I do agree it would be better to throw a warning if the user does this though.

chenghao-intel:
I've fixed this and pass the regression test.

rxin:
Thanks - do you have any idea about the performance impact of this change for joins?

chenghao-intel:
oh, I didn't do the performance test, I can do that when I back from vacation (Sept, 9th). And I also put TODOs in the code comment, where can be improved for performance.  I was planning send another pull request to resolve them.

chenghao-intel:
I just ran a join test on a real cluster, with and w/o this patch, basically there is no big difference in performance(168 s and 172 s respectively), cause the performance bottleneck are cpu, network and the disk IO alternatively. 

PS:
The HQL is: select a.pageurl, b.pageurl, c.pageurl from aa a left outer join aa b on (a.pageurl = b.pageurl and a.avgduration < 90 and a.avgduration > 0) right outer join aa c on (a.pageurl = c.pageurl and c.pagerank > 0 and c.pagerank < 5000);
And table aa is constituted of 96 files (totally 4GB snappy compressed binary columnar based).

src/main/scala/shark/execution/JoinUtil.scala
```diff
@@ -56,15 +56,18 @@ object JoinUtil {
...
 
     val isFiltered: Boolean = {
-      Range(0, filters.size()).exists { x =>
-        val cond = filters.get(x).evaluate(row)
-        val result = Option[AnyRef](
-          filtersOI.get(x).asInstanceOf[PrimitiveOI].getPrimitiveJavaObject(cond))
-        result match {
-          case Some(u) => u.asInstanceOf[Boolean].unary_!
-          case None => true
+      if(filters == null)
```
rxin: Add a space after if, and put curly braces around the body. Also, this covers more than your change, but can you rewrite the Range(0, filters.size()) this in function in while loops?

src/main/scala/shark/execution/CommonJoinOperator.scala
```diff
@@ -198,10 +198,12 @@ object CommonJoinOperator {
   val LEFT_SEMI_JOIN = JoinDesc.LEFT_SEMI_JOIN
 
   /**
-   * Handles join filters in Hive. It is kind of buggy and not used at the moment.
+   * Handles join filters in Hive
    */
   def isFiltered(row: Any, filters: JavaList[ExprNodeEvaluator], ois: JavaList[ObjectInspector])
   : Boolean = {
+    if(row == null) return true
```
rxin: add a space after if, and wrap return true in curly braces. i.e.

```
if (row == null) {
  return true
}
```

src/main/scala/shark/execution/JoinOperator.scala
```diff
@@ -139,47 +141,229 @@ class JoinOperator extends CommonJoinOperator[JoinDesc, HiveJoinOperator]
   }
 
   def generateTuples(iter: Iterator[Array[Any]]): Iterator[_] = {
-    //val tupleOrder = CommonJoinOperator.computeTupleOrder(joinConditions)
+    new TupleIterator(iter)
```
rxin: Let's just remove this generateTuples function.

chenghao-intel:
Hi, rxin, I have updated the source code, can you check that again? I will pull another request for
1) Replace the input elements from Array[Byte] with [Object], to avoid extra de-serialiazations.
2) MapJoinOperator has the same bug, but it may not has full outer join, even right outer join, I will figure out soon.

By the way, there was an Serialization Exception that I spent couple of days to debug, says the JoinDesc can not be serialized. Eventually I realized that I got confused with the Operator serialization thru class OperatorSerializationWrapper and RDD closure serialization thru spark internal implementation. Any Operator instance/property reference within RDD.mapPartitions or RDD.mapPartitionWithIndex, should always use the  OperatorSerializationWrapper instance, otherwise, the spark will try to serialize the operator itself, cause it's a closure variable.

And yes that JoinDesc can be serialized by OperatorSerializationWrapper, even it has the properties without implementing the java.io.Serializable, but spark closure serialization works differently.

src/main/scala/shark/execution/JoinOperator.scala
chenghao-intel: The class TupleIterator is the inner class of the JoinOperator, and it's not static class, so the caller can not instantiate TupleIterator directly.

rxin:
OperatorSerializationWrapper actually uses Kryo to serialize the operator. That's why it can serialize the operators even though the internal structures are not Java serializable.

rxin:
@marmbrus can you take a look at this too?

chenghao-intel:
Yes, true, I just changed the "conf" to "op.conf" within cogrouped.mapPartitions {.....}. and then it works. But it's very hard to figure out in the first glance of the code, I am wondering if we can do something.

src/main/scala/shark/execution/JoinOperator.scala
marmbrus: You can still instantiate inner classes directly given an instance of the outer class.

```
scala> class X { class Y }
defined class X
scala> val x = new X
x: X = X@22639b4a
scala> val y = new x.Y
y: x.Y = X$Y@73607ba9
```

src/main/scala/shark/execution/JoinOperator.scala
```diff
@@ -139,47 +141,229 @@ class JoinOperator extends CommonJoinOperator[JoinDesc] with ReduceSinkTableDesc
...
+        }
+      }
+    }
+
+    /**
+     * Deserialize the value,.
+     */
+    private def deserColumns(element: Array[Byte], deser: Deserializer): Array[java.lang.Object] = {
+      // TODO should reuse the kv object
+      var kv = new Array[Object](2)
```
marmbrus: I think this can be a val.

src/main/scala/shark/execution/JoinOperator.scala
```diff
@@ -139,47 +141,229 @@ class JoinOperator extends CommonJoinOperator[JoinDesc] with ReduceSinkTableDesc
...
+      if (element != null) {
+        //reuseByte.set(element, 0, element.length)
```
marmbrus: Remove dead code.

src/main/scala/shark/execution/JoinOperator.scala
```diff
@@ -139,47 +141,229 @@ class JoinOperator extends CommonJoinOperator[JoinDesc] with ReduceSinkTableDesc
...
+        var bytes = new BytesWritable(element)
```
marmbrus: Can also be val.

marmbrus:
I took a look, and this implementation looks more correct than before.  However, I think there are still a few issues with the outer join cases.  Specifically, if I understand outer join SQL semantics correctly, rows from either side that do not match the join conditions should be included only once in the output (with the other sides values set to NULL).  I think this implementation leads to over inclusion.

Consider the following test case where the input to each side of the join is (200, 200, 97, 97) and (97, 97) respectively:

``` sql
shark> CREATE TABLE src(key INT, value STRING);
shark> LOAD DATA LOCAL INPATH '${env:HIVE_HOME}/examples/files/kv1.txt' INTO TABLE src;

shark> SELECT * FROM                                   
     >   (SELECT key FROM src WHERE key = 200 OR key = 97) a
     > LEFT OUTER JOIN
     >   (SELECT key FROM src WHERE key = 97) b
     > ON a.key = 200;
OK
200 97
200 97
97  NULL
97  NULL
200 97
200 97
97  NULL
97  NULL
Time taken: 3.54 seconds
```

The non-matching 97 rows are included twice each.  In contrast, SQLite returns the following:

``` sql
sqlite> SELECT * FROM
   ...>   (SELECT 200 as key UNION ALL SELECT 200 as key UNION ALL SELECT 97 as key UNION ALL SELECT 97 as key) a
   ...> LEFT OUTER JOIN
   ...>   (SELECT 97 as key UNION ALL SELECT 97 as key) b
   ...> ON a.key = 200;
200|97
200|97
200|97
200|97
97|
97|
```

src/main/scala/shark/execution/JoinOperator.scala
```diff
@@ -139,47 +141,229 @@ class JoinOperator extends CommonJoinOperator[JoinDesc] with ReduceSinkTableDesc
...
+    new TupleIterator(iter)
+  }
 
-    // TODO: use MutableBytesWritable to avoid the array copy.
-    val bytes = new BytesWritable
-    val tmp = new Array[Object](2)
+  override def processPartition(split: Int, iter: Iterator[_]): Iterator[_] =
+    throw new UnsupportedOperationException("JoinOperator.processPartition()")
 
+  class TupleIterator(iter: Iterator[Array[Any]]) extends Iterator[Object] {
```
marmbrus: Can you document the assumptions about the input to `TupleIterator` and what it produces?

chenghao-intel:
Thanks @marmbrus , I tried the same query on Hive, it produced the same result with Shark, but MySQL worked exactly the same with SQLite. Your understanding should be correct, I will figure out how to fix it.

chenghao-intel:
@marmbrus @rxin , I've updated the code:
1) Add JoinFilter support for MapJoin
2) Improve the performance for JoinOperator, to avoid multiple times of deserialization.
3) Fix ClassCastException in MapJoin (NullWritable cannot be casted into xxxWritable)
4) Add some documentation

Last thing is about the "Null should be included only once if either side doesn't match the join condition", I am wondering if we can keep the same logical with Hive. Or I can create another pull request to fix that.

BTW: I merged several commits into single one, which may be more convenient in code review. Or should I keep it as several commits in the future? 

Thanks,
Cheng Hao

ss77892:
Could you please explain in two words what have been done with 3th bullet?
Have you added the handling for NullWritable or just removed unnecessary
serialization from MapJoin?

Thanks,
Sergey

chenghao-intel:
In MapJoin, the small table values will be wrap as SerializableWritable (see SerializableWritable.scala), the corner case is the null value in the small tables, which is transformed to NullWritable in serialization, but during later de-serialization, it supposedly to be null again, not NullWritable, which is the bug in previous implementation.

Please review the file SerializableWritable.scala of this PR. Sorry, I should make that as a single commit.

src/main/scala/shark/execution/JoinFilter.scala
```diff
@@ -0,0 +1,283 @@
...
+  
+  /**
+   * check the join filter (true for discarding the data)
+   */
+  protected def isFiltered(tableData: AnyRef, index: Byte): Boolean = {
+    if (noOuterJoin || tableData == null) {
+      false
+    } else {
+      val columns = tableData.asInstanceOf[Array[AnyRef]]
+      columns(columns.length - 1).asInstanceOf[BooleanWritable].get()
```
harveyfeng: This section looks hardcoded...could you add more detailed documentation here? It also looks like CommonJoinFilter#isFiltered() isn't used.

harveyfeng: There's also a duplicate filter helper method in Hive's JoinUtil.java. How are the exec. paths different from that version?

harveyfeng:
Hey @chenghao-intel,
We cherry-picked a previous version of this patch (see 6fc3d3f), from this PR: https://github.com/amplab/shark/pull/250 into this one: https://github.com/amplab/shark/pull/259

I tried merging in your current changes and resolving a few conflicts, but got ClassCast runtime exceptions caused by `JoinFilter#isFiltered()` :\
If you have the change to rebase this against the master branch, then we can include it in the next v0.9 minor release. Please LMK if there's anything we can do to help!

chenghao-intel:
Thanks @harveyfeng , I will do that.

chenghao-intel:
@harveyfeng , I have merged with master, and fixed the ClassCast exception.
But I also noticed Hive 0.11 has fixed the outer join filter mentioned by @marmbrus in above discussion, hence I updated this join filter implementation accordingly.

harveyfeng:
Thanks for doing that, and the clean documentation! We'll take a closer look at it this in a bit

rxin:
Jenkins, retest this please.

chenghao-intel:
Updated the code, tests_pass.txt and tests_fail.txt, it passed the unit test in my local. Can you verify it again?

I will send out the join performance test in real cluster later.

chenghao-intel:
I've finished the performance comparison before and after the patch,
SQL   OLD(secs)  NEW (secs)
1         96.3         24.27
2         144          21.398
3         96.1         22.367
4         93.18           24.103

The new implementation does improve a lot in performance (4x ~ 6x).

Below is the SQL that I used in the testing.
1) SELECT sourceIP, pageRank, adRevenue FROM rankings R inner join uservisits UV on R.pageURL = UV.destURL limit 1;

2) SELECT sourceIP, pageRank, adRevenue FROM rankings R right outer join uservisits UV on R.pageURL = UV.destURL and UV.sourceIP > '0' and UV.sourceIp < 'zz' limit 1;

3) SELECT sourceIP, pageRank, adRevenue FROM rankings R left outer join uservisits UV on R.pageURL = UV.destURL and UV.sourceIP > '0' and UV.sourceIp < 'zz' limit 1;

4) SELECT sourceIP, pageRank, adRevenue FROM rankings R full outer join uservisits UV on R.pageURL = UV.destURL and UV.sourceIP > '0' and UV.sourceIp < 'zz' limit 1;

rxin:
Jenkins, retest this please.

harveyfeng:
I've reverted the _cached PR that introduced some noise in the Jenkins runs and updated master to use the latest Hive 0.11 SNAPSHOT jars. You'll have to merge in those changes to Jenkins to run smoothly for this PR.

Sorry about that hassle :\

rxin:
Jenkins, retest this please.

PR #163, opened 2013-09-16
Pull the Hadoop version from the environment.

sarahgerweck:
This introduces two environment variables that you can use to control
the version of Hadoop that you build against. The Spark project
uses `SPARK_HADOOP_VERSION`, which is supported. By analogy, we
also support `SHARK_HADOOP_VERSION`, which has precedence over
`SPARK_HADOOP_VERSION`. If neither of these environment variables are
set at build time, the defaults from the build script are used.

The algorithm looks like this:
1. Use `SHARK_HADOOP_VERSION` if set.
2. Use `SPARK_HADOOP_VERSION` if set.
3. Use the default that is hard-coded into the build definition.

### DIFF
--- a/project/SharkBuild.scala
+++ b/project/SharkBuild.scala
@@ -32,7 +32,13 @@ object SharkBuild extends Build {
 
   // Hadoop version to build against. For example, "0.20.2", "0.20.205.0", or
   // "1.0.1" for Apache releases, or "0.20.2-cdh3u3" for Cloudera Hadoop.
-  val HADOOP_VERSION = "1.0.4"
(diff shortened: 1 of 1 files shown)


rxin:
Jenkins, test this please.

sarahgerweck:
You don't need anything from me for this do you? The change is very conservative.

rxin:
Yea - the changes look good. I was waiting for Jenkins to respond but something is wrong with it.

sarahgerweck:
Cool, just making sure. :-)

PR #165, opened 2013-09-18
Fix hadoop conf dir path for Hadoop 2.x

colorant:
Try to locate correct path for hadoop conf dir depending on the version.

### DIFF
--- a/run
+++ b/run
@@ -106,9 +106,10 @@ SPARK_CLASSPATH+=":$SHARK_HOME/target/scala-$SCALA_VERSION/test-classes"
 
 
 if [ "x$HADOOP_HOME" == "x" ] ; then
-    echo "No HADOOP_HOME specified. Shark will run in local-mode"
+  echo "No HADOOP_HOME specified. Shark will run in local-mode"
 else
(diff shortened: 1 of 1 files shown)


run
```diff
@@ -106,9 +106,14 @@ SPARK_CLASSPATH+=":$SHARK_HOME/target/scala-$SCALA_VERSION/test-classes"
 
 
 if [ "x$HADOOP_HOME" == "x" ] ; then
-    echo "No HADOOP_HOME specified. Shark will run in local-mode"
+  echo "No HADOOP_HOME specified. Shark will run in local-mode"
 else
+  HADOOP_EXECUTABLE=$HADOOP_HOME/bin/hadoop
```
pwendell: Any reason not to just add both $HADOOP_HOME/etc/hadoop and $HADOOP_HOME/conf every time? In my experience the hadoop scripts can look different in different distributions, it might be good to just set both rather than relying on anything specific to the format of the scripts.

colorant:
@pwendell sounds reasonable, thanks. updated.

pwendell:
@rxin Looks good to me

rxin:
Thanks.

PR #167, opened 2013-09-18
Took log4j init out of SharkContext.

rxin:
This way, applications that use SharkContext can properly initialize their own log4j configuration.


rxin:
Jenkins, test this please.

PR #168, opened 2013-09-18
Correcting Spark dependency to 0.8-incubating-SNAPSHOT

ghost:
Spark has new versions for the artifact and Shark build is broken because of it.
Let's update. 

Same fix goes for master branch.


ghost:
Now this commit is updated with the implementation of SHARK-120.

rxin:
Thanks. Merging this one for now, but FYI I am not sure if we want to maintain 2 build systems in the long term. I'd follow Spark's decision in the end to focus on one, either SBT or Maven.

ghost:
Thanks Reynold, appreciate the merge. I agree that two build systems are a PITA - let's make the decision once we are ready for it.

PR #172, opened 2013-09-20
SHARK-183: implementing BooleanBitSetCompression

AndreSchumacher:
Added an encoder for Booleans into Bytes to the Memstore that uses the CompressionAlgorithm API. Thanks Reynold for help and discussion.

### DIFF
--- a/src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
+++ b/src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
@@ -19,6 +19,8 @@ package shark.memstore2.column
 
 import java.nio.ByteBuffer
 
+import org.apache.hadoop.io.BooleanWritable
+
 import shark.memstore2.column.Implicits._
 
 /**
@@ -41,6 +43,7 @@ trait CompressedColumnIterator extends ColumnIterator {
       case DefaultCompressionType => new DefaultDecoder(buffer, columnType)
       case RLECompressionType => new RLDecoder(buffer, columnType)
       case DictionaryCompressionType => new DictDecoder(buffer, columnType)
+      case BooleanBitSetCompressionType => new BooleanBitSetDecoder(buffer, columnType)
       case _ => throw new UnsupportedOperationException()
     }
   }
@@ -123,3 +126,35 @@ class DictDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends I
     _dictionary(index)
   }
 }
+
+/**
+ * Boolean BitSet encoding.
+ */
+class BooleanBitSetDecoder[V](
+    buffer: ByteBuffer,
+    columnType: ColumnType[_, V],
+    var _pos: Int,
+    var _uncompressedSize: Int,
+    var _curValue: Long,
+    var _writable: BooleanWritable
+  ) extends Iterator[V] {
+
+  def this(buffer: ByteBuffer, columnType: ColumnType[_, V])
+      = this(buffer, columnType, 0, buffer.getInt(), 0, new BooleanWritable())
+
+  override def hasNext = _pos < _uncompressedSize
+
+  override def next(): V = {
+    val offset = _pos % BooleanBitSetCompression.BOOLEANS_PER_LONG
+
+    if (offset == 0) {
+      _curValue = buffer.getLong()
+    }
+
+    val retval: Boolean = (_curValue & (1 << offset)) != 0
+    _pos += 1
+    _writable.set(retval)
+    _writable.asInstanceOf[V]
+  }
+}
+
--- a/src/main/scala/shark/memstore2/column/CompressionAlgorithm.scala
+++ b/src/main/scala/shark/memstore2/column/CompressionAlgorithm.scala
@@ -53,6 +53,8 @@ object RLECompressionType extends CompressionType(0)
 
 object DictionaryCompressionType extends CompressionType(1)
 
+object BooleanBitSetCompressionType extends CompressionType(2)
+
 /**
  * An no-op compression.
  */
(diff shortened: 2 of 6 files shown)


rxin:
Jenkins, test this please.

src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
```diff
@@ -123,3 +126,36 @@ class DictDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends I
...
+class BooleanBitSetDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends Iterator[V] {
+
+  // The position in the current Byte.
+  private var _bitPos : Int = -1
+
+  // The number of bits in the next-to-last Byte (contained in last Byte).
+  private var _lastBitIndex : Int = _
+
+  // The current Byte value.
+  private var _curByte: Byte = _
```
rxin: To better align this, let's read one word (Long) at a time instead of one byte at a time.

src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
```diff
@@ -123,3 +126,36 @@ class DictDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends I
...
+/**
+ * Boolean BitSet encoding.
+ */
+class BooleanBitSetDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends Iterator[V] {
+
+  // The position in the current Byte.
+  private var _bitPos : Int = -1
+
+  // The number of bits in the next-to-last Byte (contained in last Byte).
+  private var _lastBitIndex : Int = _
```
rxin: Why don't we just save the length in the first word? That can simplify your logic.

src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
```diff
@@ -123,3 +126,36 @@ class DictDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends I
...
   }
 }
+
+/**
+ * Boolean BitSet encoding.
+ */
+class BooleanBitSetDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends Iterator[V] {
+
+  // The position in the current Byte.
+  private var _bitPos : Int = -1
```
rxin: Our coding style doesn't add the whitespace before the colon, i.e.

```
private var _bitPos: Int = -1
```

src/main/scala/shark/memstore2/column/CompressionAlgorithm.scala
```diff
@@ -277,3 +279,76 @@ class DictionaryEncoding extends CompressionAlgorithm {
...
+
+    var cur : Byte = 0
+    var pos : Int = 0
+    var offset : Int = 0
+
+    while(b.hasRemaining) {
+      offset = pos % BooleanBitSetCompression.BOOLEANS_PER_BYTE
+      val elem = t.extract(b).asInstanceOf[Boolean]
+
+      if(elem) {
```
rxin: Add whitespace after "if"

rxin:
Hi Andre,

This looks pretty good overall. I think the logic can be a little bit simpler if we just write out the length in the beginning of the buffer. Also made a couple comments on coding style.

AndreSchumacher:
Thanks. Good point about just saving the length at the beginning. Guess I was still thinking of iterators when I wrote this. Will fix these issues and then update the PR.

AndreSchumacher:
So I did the changes. Strange that Jenkins doesn't want to pull it though.

src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
```diff
@@ -123,3 +126,38 @@ class DictDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends I
...
+  // The number of Booleans in uncompressed buffer.
+  private var _uncompressedSize: Int = _
+
+  // The current record (Long) value.
+  private var _curValue: Long = _
+
+  override def hasNext = _pos < _uncompressedSize
+
+  override def next(): V = {
+    if (_pos == -1) { // this is the first Boolean to be read
```
rxin: Move this to the constructor so we can skip an if branch in next()

src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
```diff
@@ -123,3 +126,38 @@ class DictDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends I
...
+
+    val offset = _pos % BooleanBitSetCompression.BOOLEANS_PER_LONG
+
+    if (offset == 0) {
+      _curValue = buffer.getLong()
+    }
+
+    val retval: Boolean = (_curValue & (1 << offset)) != 0
+    _pos += 1
+    new BooleanWritable(retval).asInstanceOf[V]
```
rxin: Reuse an existing boolean writable

src/main/scala/shark/memstore2/column/CompressionAlgorithm.scala
```diff
@@ -277,3 +279,71 @@ class DictionaryEncoding extends CompressionAlgorithm {
...
+  override def compressionType = BooleanBitSetCompressionType
+
+  override def supportsType(t: ColumnType[_, _]) = {
+    t match {
+      case BOOLEAN => true
+      case _ => false
+    }
+  }
+
+  override def gatherStatsForCompressibility[T](v: T, t: ColumnType[T,_]) {
```
rxin, line 303: This function can be empty, and just change compressedSize to

```
override def compressedSize: Int = Math.ceil(_uncompressedSize.toFloat / 64) * 8 + 4
```

src/test/scala/shark/memstore2/column/CompressionAlgorithmSuite.scala
```diff
@@ -258,4 +258,34 @@ class CompressionAlgorithmSuite extends FunSuite {
...
+      bbs.gatherStatsForCompressibility(false, BOOLEAN)
+    }
+    b.put(0.toByte)
+    b.put(1.toByte)
+    bbs.gatherStatsForCompressibility(false, BOOLEAN)
+    bbs.gatherStatsForCompressibility(true, BOOLEAN)
+    b.limit(b.position())
+    b.rewind()
+    val compressedBuffer = bbs.compress(b, BOOLEAN)
+    assert(compressedBuffer.getInt() == BOOLEAN.typeID)
```
rxin: For assert statements, use === instead of ==.

=== will actually print out a more meaningful message to tell you value x != value y when the assert fails.

rxin:
Thanks. One more round ... Jenkins is kind of buggy for Shark right now.

src/main/scala/shark/memstore2/column/CompressionAlgorithm.scala
AndreSchumacher, line 303: Isn't the problem that the compression algorithm is chosen at runtime depending on which one yields the smaller compressed data size? So at the time when compressedSize (or uncompressedSize) is called the algorithm hasn't even seen the uncompressed byte buffer and only way to know is via this gatherStatsForCompressability method.

I agree the current version seems wasteful but unless one changes the "contract" between the compression and the surrounding code I see no other solution but please correct me if I missed something.

AndreSchumacher:
Jenkins, test this please.

src/main/scala/shark/memstore2/column/CompressionAlgorithm.scala
rxin, line 303: What I meant was this function can just calculate the uncompressedSize (which is basically incrementing by 1 byte per element), and then compressedSize is imply uncompressedSize / 8 with rounding.

AndreSchumacher, line 303: Ah, OK, I thought you suggested to get rid of gatherStatsForCompressibility completely. Sure, will change that.

rxin:
Jenkins, retest this please.

AndreSchumacher:
The Jenkins console output looks quite different from what I get when I run "sbt/sbt test". Does Jenkins actually run the Shark unit tests?

harveyfeng:
It looks like Jenkins uses bin/dev/run-tests-from-scratch to run the Shark unit tests first, followed by the Hive unit test suite. Most of the output is from Hive unit tests - Shark test output can be found somewhere after the search result for "sbt/sbt clean test" in the raw output.

PR #174, opened 2013-09-22
Do not init spark context when in remote mode

colorant:
when running shark cli to connect to shark server, there are no need to launch a second spark context since all cmd is go through the server's context.


rxin:
Jenkins, test this please.

rxin:
Thanks, Raymond!

PR #175, opened 2013-09-25
Support for caching Hive-partitioned tables, including support for more DDL statements on cached tables.

harveyfeng:
- Added a SparkDDLTask that is dependent on Hive's DDLTask.
  - Makes it possible to create cached tables (partitioned or not) using a regular CREATE TABLE command, e.g. `CREATE TABLE test_cached (key string, value string)`
  - With additions to SharkDDLSemanticAnalyzer, it should be possible to support more DDL commands, notably ALTER. I haven't added ones such as table drop protections and ALTER TABLE ... RENAME TO .... , but it should be simple to do in a subsequent PR.
- Didn't include support for partition pruning on RDDs representing Hive-partitions - will add that in another PR.


harveyfeng:
PTAL at the core functionalities added - will update with more comments, and possibly small style changes, throughout this week

Thanks!

harveyfeng:
Will reopen once an eviction policy is added, plus some more comments.

PR #176, opened 2013-09-25
Bug fix in partition/map pruning for between clauses.

rxin:


rxin:
Jenkins, test this please.

PR #179, opened 2013-09-28
Delta Encoding using a single byte for small deltas now a compression sc...

junkumar:
...heme.


rxin:
Jenkins, add to whitelist.

rxin:
Oops - accidentally closed this one.

src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
```diff
@@ -131,16 +137,15 @@ class DictDecoder[V](buffer: ByteBuffer, columnType: ColumnType[_, V]) extends I
...
  */
 class BooleanBitSetDecoder[V](
-    buffer: ByteBuffer,
-    columnType: ColumnType[_, V],
-    var _pos: Int,
-    var _uncompressedSize: Int,
-    var _curValue: Long,
-    var _writable: BooleanWritable
-  ) extends Iterator[V] {
+  buffer: ByteBuffer,
```
rxin: Can we change this style back to the original one (4 space indent for function/class parameters)?

src/main/scala/shark/memstore2/column/CompressionAlgorithm.scala
```diff
@@ -145,7 +165,7 @@ class RLE extends CompressionAlgorithm {
   }
 
   @tailrec private final def encode[T](currentBuffer: ByteBuffer,
-      compressedBuffer: ByteBuffer, currentRun: (T, Int), t: ColumnType[T,_]) {
+    compressedBuffer: ByteBuffer, currentRun: (T, Int), t: ColumnType[T, _]) {
```
rxin: roll back to the 4 space indent in function definition.

src/main/scala/shark/memstore2/column/ColumnBuilder.scala
```diff
@@ -145,7 +160,13 @@ object ColumnBuilder {
       case _ => new GenericColumnBuilder(columnOi)
     }
     if (shouldCompress) {
-      v.compressionSchemes = Seq(new RLE(), new BooleanBitSetCompression())
+      v.compressionSchemes = Seq(new NoCompression,
+        new RLE,
+        new BooleanBitSetCompression,
+        new DictionaryEncoding)
+      if(bde != null) {
```
rxin: Why do we need this bde here given the bde itself defines it only supports long, short, and int?

src/main/scala/shark/memstore2/column/CompressionAlgorithm.scala
```diff
@@ -336,7 +358,116 @@ class BooleanBitSetCompression extends CompressionAlgorithm {
...
+  * Delta encoding for numeric columns. This algorithm encodes values into a single byte whenever
+  * the difference from the previous value (Delta) is small enough.
+  * A flag byte is always used for each value.
+  *
+  * The first bit of that flag byte acts as an escape code that identifies whether the next few
+  * bytes (fixed number based on Numeric Type) should be used or not.
+  *  0 implies there is a new baseValue in the following byte
+  *  1 implies that the delta is encoded into the next seven bits
+  *
+  * The best case is a small difference from previous value - small enough to fit in 7 bits - a
```
rxin, line 381: In addition to this, can you give some examples to illustrate how this works? For example, you can use 1, 2, 3, 5, 100, 5000 (or some other numbers that are easier to represent in bits)

src/main/scala/shark/memstore2/column/ColumnBuilder.scala
junkumar: new ByteDeltaEncoding[T] needs to be initialized with the T. This is why bde is initialized differently than the other types. I can only support Numeric types because I have to be able to compute deltas unlike the other compression schemes so far.

junkumar:
Reynold,

Is something wrong with the paths for sbt?

The build is failing with
- SPARK_WITH_YARN=false
- sbt/sbt clean publish-local
  ./bin/dev/run-tests-from-scratch: line 254: sbt/sbt: No such file or directory
  Build step 'Execute shell' marked build as failure
  Finished: FAILURE

rxin:
Jenkins, retest this please.

rxin:
Hi Nandu,

I looked into the generated code for Numeric plus. It actually doesn't invoke -. Rather, it is just an implicit that calls methods defined in Numerics. It actually involves boxing/unboxing and is much slower than the arithmetic. 

For example, for the following code:

```
def plus[N: Numeric](x: N, y: N): N = x - y
println(plus(1, 2))
```

The scala compiler generates: 

```
def plus(x: Object, y: Object, evidence$1: math.Numeric): Object = scala.math.Numeric$Implicits.infixNumericOps(x, evidence$1).-(y);
Test.this.plus(scala.Int.box(1), scala.Int.box(2), scala.math.Numeric$IntIsIntegral)
```

I am afraid you will have to do something similar to what you have done for the decoder (which is really cool).

src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
```diff
@@ -158,3 +163,73 @@ class BooleanBitSetDecoder[V](
...
+          prev.asInstanceOf[LongWritable].set(pvalue.toLong)
+          prev
+        }
+      }
+      case _ => throw new UnsupportedOperationException("Unsupported data type " + columnType)
+    }
+  }
+
+  override def next(): W = {
+    BYTE.extractInto(buffer, byteWritable)
```
rxin: You can replace these two lines with simply

```
flagByte = buffer.get()
```

and remove byteWritable variable from the class.

src/main/scala/shark/memstore2/column/CompressedColumnIterator.scala
```diff
@@ -158,3 +163,73 @@ class BooleanBitSetDecoder[V](
...
+  var prev: W = columnType.newWritable()
+  var startedDecoding = false
+  var flagByte: Byte = _
+
+  private val current: W = columnType.newWritable()
+  private val byteWritable = BYTE.newWritable()
+
+  override def hasNext = buffer.hasRemaining
+
+  def plus[N: Numeric](x: N, y: N): N = x + y
```
rxin: This plus is no longer used.

PR #180, opened 2013-09-29
Adding startup scripts to the maven assembly (missed at first run)

ghost:
Adding configs and scripts to the maven assembly.


rxin:
Thanks.

PR #181, opened 2013-10-05
Add a SparkDDLTask for handling Shark-specific metastore operations.

harveyfeng:
Includes support for ALTER TABLE <old naem> RENAME TO <new name> on cached tables.

Currently, this is meant to be a dependent task on Hive's DDLTask.

PR #182, opened 2013-10-07
Support for in-memory, Hive-partitioned tables.

harveyfeng:
Note: dependent on https://github.com/amplab/shark/pull/181

Added along the way:
- Creating tables without CTAS - e.g., CREATE TABLE test_cached (...) [PARTITIONED BY (...)]
- General cache policy interface and LRU implementation, since there isn't a good way to specify, from the CLI, which partitions to keep in memory.
  Table class, with MemoryTable and PartitionedMemoryTable subclasses, for containing RDD metadata.

Partition pruning and stats collection is still unsupported/untested. Also, some additions made a few files, especially MemoryStoreSinkOperator, more confusing. Planning to submit fixes for these two issues in another CL, since it might involve more refactoring.


src/main/scala/shark/SharkConfVars.scala
```diff
@@ -43,6 +43,23 @@ object SharkConfVars {
   // Default storage level for cached tables.
   val STORAGE_LEVEL = new ConfVar("shark.cache.storageLevel", "MEMORY_AND_DISK")
 
+  // True if a table's partitions should be evicted using a cache policy.
+  val SHOULD_USE_CACHE_POLICY = new ConfVar("shark.cache.partition.useCachePolicy", true)
+
+  // Class name of the default cache policy used to manage partition evictions for cached,
+  // Hive-partitioned tables.
+  val CACHE_POLICY = new ConfVar(
+    "shark.cache.partition.cachePolicy.class", "shark.memstore2.LRUCachePolicy")
```
rxin: let's rename this to

shark.partition.cachePolicy.default

rxin: or 

shark.cache.partition.policy

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -38,12 +38,36 @@ import shark.tachyon.TachyonTableWriter
  */
 class MemoryStoreSinkOperator extends TerminalOperator {
 
+  // The initial capacity for ArrayLists used to construct the columnar storage. If -1,
+  // the ColumnarSerde will obtain the partition size from a Configuration during execution
+  // initialization (see ColumnarSerde#initialize()).
   @BeanProperty var partitionSize: Int = _
```
rxin, line 44: This is not about your change, but let's get rid of the shark.column.partitionSize.mb setting, and all those crazy heuristics to estimate column buffer's size. Let's just use 1MB as the initial size for a column. It is pretty fast to do a mem copy anyway. It would simplify some code.

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -74,36 +98,42 @@ class MemoryStoreSinkOperator extends TerminalOperator {
...
-      val serde = new ColumnarSerDe
-      serde.initialize(op.localHconf, op.localHiveOp.getConf.getTableInfo.getProperties)
-
-      // Serialize each row into the builder object.
-      // ColumnarSerDe will return a TablePartitionBuilder.
-      var builder: Writable = null
-      iter.foreach { row =>
-        builder = serde.serialize(row.asInstanceOf[AnyRef], op.objectInspector)
-      }
+    var outputRDD: RDD[TablePartition] = inputRdd.mapPartitionsWithIndex {
```
rxin: Can we move case(partitionIndex, iter) to the previous line, and rename partitionIndex to part? Then we don't have to do the extra indents.

src/main/scala/shark/execution/MemoryStoreSinkOperator.scala
```diff
@@ -122,21 +153,53 @@ class MemoryStoreSinkOperator extends TerminalOperator {
...
+      }
+    } else {
+      outputRDD.setName(tableName)
+      // Create a new MemoryTable entry if one doesn't exist (i.e., this operator is for a CTAS).
+      val memoryTable = SharkEnv.memoryMetadataManager.getMemoryTable(tableName).getOrElse(
+        SharkEnv.memoryMetadataManager.createMemoryTable(tableName, cacheMode, storageLevel))
+      memoryTable.tableRDD = outputRDD
     }
 
     // Report remaining memory.
```
rxin: Let's remove the memory reporting code. The Spark web ui is much better at reporting that.

src/main/scala/shark/execution/RDDUtils.scala
```diff
@@ -34,22 +34,56 @@ import shark.SharkEnv
...
     }
   }
 
   /**
+   * Returns the storage level of a sequence of RDDs, interpreted as the storage level of the first
+   * RDD in the sequence that persisted in memory or disk.
+   *
+   * @param rdds The sequence of RDDs to find the StorageLevel of.
+   */
+  def getStorageLevelOfRDDs(rdds: Seq[RDD[_]]): StorageLevel = {
```
rxin: In our use case, is it possible for these rdds to have different storage levels?

src/main/scala/shark/execution/RDDUtils.scala
```diff
@@ -34,22 +34,56 @@ import shark.SharkEnv
...
+    rdds.foldLeft(StorageLevel.NONE) {
+      (s, r) => {
+        if (s == StorageLevel.NONE) {
+          // Mutally recursive if 'r' is a UnionRDD.
```
rxin: Why does this happen? (nested union rdd?)

src/main/scala/shark/execution/SharkDDLTask.scala
```diff
@@ -25,45 +27,149 @@ import org.apache.hadoop.hive.ql.metadata.Hive
...
+ * valid one.
+ */
+private[shark] class SharkDDLTask extends HiveTask[SharkDDLWork]
+  with Serializable with LogHelper {
 
   override def execute(driverContext: DriverContext): Int = {
-    val hiveMetadataDb = Hive.get(conf)
-
     work.ddlDesc match {
+      case creatTblDesc: CreateTableDesc => {
```
rxin: for the body of case, you can remove the { } so it is slightly shorter

src/main/scala/shark/execution/SharkDDLTask.scala
```diff
@@ -25,45 +27,149 @@ import org.apache.hadoop.hive.ql.metadata.Hive
...
 
-  def alterTable(
-      hiveMetadataDb: Hive,
-      alterTableDesc: AlterTableDesc) {
-    if (alterTableDesc.getOp() == AlterTableDesc.AlterTableTypes.RENAME) {
-      val oldName = alterTableDesc.getOldName
-      val newName = alterTableDesc.getNewName
-      SharkEnv.memoryMetadataManager.rename(oldName, newName)
+  /** Handles a CREATE TABLE or CTAS. */
+  def createTable(
```
rxin: Put the function declaration on 1 line instead of 3

src/main/scala/shark/execution/SharkDDLTask.scala
```diff
@@ -25,45 +27,149 @@ import org.apache.hadoop.hive.ql.metadata.Hive
...
+    if (partitionSpecs == null) {
+      // The command is a true DROP TABLE.
+      SharkEnv.dropTable(tableName)
+    } else {
+      // The command is an ALTER TABLE DROP PARTITION
+      val partitionedTable = getPartitionedTableWithAssertions(tableName)
+      // Find the set of partition column values that specifies the partition being dropped.
+      val partitionColumns: Seq[String] = hiveTable.getPartCols.map(_.getName)
+      for (partitionSpec <- partitionSpecs) {
+        val partitionColumnToValue: JavaMap[String, String] =
```
rxin: Why don't you rename JavaMap to JMap and then u can squeeze this in one line :)

rxin: or rename partitionColumns / partitionColumnToValue to be slightly shorter...

src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -239,17 +247,67 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
...
+      val partDesc = Utilities.getPartitionDesc(partition)
+      // Get partition field info
+      val partSpec = partDesc.getPartSpec()
+      val partProps = partDesc.getProperties()
+
+      val partColsDelimited = partProps.getProperty(META_TABLE_PARTITION_COLUMNS)
+      // Partitioning keys are delimited by "/"
+      val partColumns = partColsDelimited.trim().split("/").toSeq
+      // 'partValues[i]' contains the value for the partitioning column at 'partColumns[i]'.
+      val partValues = partColumns.map { key =>
```
rxin: Can we rewrite this as

```
val partValues = if (partSpec == null) {
  Arrays.fill[partColumns.size](new String)
} else {
  ...
}
```

src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -239,17 +247,67 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
...
-   * Create an RDD for every partition column specified in the query. Note that for on-disk Hive
+   * Fetch an RDD from the Shark metastore using each partition key given, and return a union of all
+   * the fetched RDDs.
+   *
+   * @param tableKey Name of the partitioned table.
+   * @param partitions A collection of Hive-partition metadata, such as partition columns and
+   *     partition key specifications.
+   */
+  private def makeCachedPartitionRDD(tableKey: String, partitions: Array[Partition]): RDD[_] = {
+    val hivePartitionRDDSeq = partitions.map { partition =>
```
rxin: hivePartitionRDDSeq -> hivePartitionRDDs?

src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -239,17 +247,67 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
...
+          }
+        } else {
+         Iterator()
+        }
+      }
+    }
+    if (hivePartitionRDDSeq.size > 0) {
+      new UnionRDD(hivePartitionRDDSeq.head.context, hivePartitionRDDSeq)
+    } else {
+      SharkEnv.sc.makeRDD(Seq[Object]())
```
rxin: This is not your change, but make it Seq.empty[Object] to be more clear.

```
SharkEnv.sc.makeRDD(Seq.empty[Object])
```

src/main/scala/shark/execution/SharkDDLTask.scala
harveyfeng: done

src/main/scala/shark/execution/SharkDDLTask.scala
harveyfeng: done

src/main/scala/shark/execution/SharkDDLTask.scala
harveyfeng: Renamed "partition" => "part" in most of the variables

src/main/scala/shark/execution/TableScanOperator.scala
harveyfeng: Replaced this part with an EmptyRDD[Object] instantiation

src/main/scala/shark/execution/TableScanOperator.scala
harveyfeng: done

src/main/scala/shark/execution/TableScanOperator.scala
harveyfeng: Done - did the same with makeHivePartitionRDDs(). There's a bit of overlap in the two methods, though I was hoping to refactor in a following PR to make the diffs cleaner... it'd fit with some other general cleanup, such as removing EmptyRDD from shark.execution.

src/main/scala/shark/execution/RDDUtils.scala
harveyfeng: RDDs for a table should have the same storage level, though an RDD for a partitioned table could be StorageLevel.NONE if it was unpersisted by the partition eviction policy.

Clarified the comment a bit.

src/main/scala/shark/execution/RDDUtils.scala
harveyfeng: It could happen for multiple "insert into" commands on the same table.

src/main/scala/shark/memstore2/PartitionedMemoryTable.scala
```diff
@@ -0,0 +1,141 @@
...
+  // The conventional string format for the partition key, 'col1=value1/col2=value2/...', can be
+  // computed using MemoryMetadataManager#makeHivePartitionKeyStr()
+  private var _keyToPartitions: ConcurrentMap[String, RDDValue] =
+    new ConcurrentJavaHashMap[String, RDDValue]()
+
+  // The eviction policy for this table's cached Hive-partitions. An example of how this
+  // can be set from the CLI:
+  //   'TBLPROPERTIES("shark.partition.cachePolicy", "LRUCachePolicy")'.
+  // If 'None', then all partitions will be persisted in memory using the 'preferredStorageLevel'.
+  private var _cachePolicy: Option[CachePolicy[String, RDDValue]] = None
```
rxin: Let's always have a cache policy, and implement a CacheAll policy to get rid of the Option.

src/main/scala/shark/memstore2/MemoryTable.scala
```diff
@@ -0,0 +1,44 @@
...
+  extends Table(tableName, cacheMode, preferredStorageLevel) {
+
+  // RDD that contains the contents of this table.
+  private var _tableRDD: RDD[TablePartition] = _
+
+  def tableRDD: RDD[TablePartition] = _tableRDD
+
+  def tableRDD_= (rdd: RDD[TablePartition]) = _tableRDD = rdd
+
+  override def getCurrentStorageLevel: StorageLevel = RDDUtils.getStorageLevelOfRDD(tableRDD)
```
rxin: Let's remove this function since it doesn't really work for partitioned tables.

src/main/scala/shark/memstore2/LRUCachePolicy.scala
```diff
@@ -0,0 +1,94 @@
...
+  var isInitialized = false
+  var hasRecordedStats = false
+  var cache: LoadingCache[K, V] = _
+  var cacheStats: Option[CacheStats] = None
+
+  override def initialize(
+      maxSize: Long,
+      loadFunc: (K => V),
+      evictionFunc: (K, V) => Unit,
+      shouldRecordStats: Boolean
```
rxin: Let's always record stats so we can remove this variable.

src/main/scala/shark/memstore2/LRUCachePolicy.scala
```diff
@@ -0,0 +1,94 @@
...
+    super.initialize(maxSize, loadFunc, evictionFunc, shouldRecordStats)
+
+    var builder = CacheBuilder.newBuilder().maximumSize(maxSize)
+    if (shouldRecordStats) {
+      builder.recordStats()
+      hasRecordedStats = true
+    }
+
+    // Automatically executed by 'cache' on the entry being evicted.
+    val removalListener =
```
rxin: put new RemovalListener on the same line as removalListener

rxin: also do the same for CacheLoader

src/main/scala/shark/memstore2/LRUCachePolicy.scala
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package shark.memstore2
+
+import com.google.common.cache._
+
+import scala.collection.JavaConversions._
+
+
+class LRUCachePolicy[K <: AnyRef, V <: AnyRef] extends CachePolicy[K, V] {
```
rxin: We need a Least Recently Added policy

src/main/scala/shark/execution/RDDUtils.scala
```diff
@@ -34,22 +34,59 @@ import shark.SharkEnv
  */
 object RDDUtils {
 
-  def getStorageLevelOfCachedTable(rdd: RDD[_]): StorageLevel = {
+  def getStorageLevelOfRDD(rdd: RDD[_]): StorageLevel = {
```
rxin: We don't actually need this anymore.

src/main/scala/shark/memstore2/PartitionedMemoryTable.scala
```diff
@@ -0,0 +1,141 @@
...
+    ) {
+    val newPolicy = Class.forName(cachePolicyStr).newInstance
+      .asInstanceOf[CachePolicy[String, RDDValue]]
+    // The loadFunc will upgrade the persistence level of the RDD to the preferred storage level.
+    val loadFunc: String => RDDValue =
+      (partitionKey: String) => {
+        val rddValue = _keyToPartitions.get(partitionKey).get
+        if (RDDUtils.getStorageLevelOfRDD(rddValue.rdd) == StorageLevel.NONE) {
+          rddValue.rdd.persist(preferredStorageLevel)
+        }
```
harveyfeng: Todo: add comment about UnionRDDs

src/main/scala/shark/execution/RDDUtils.scala
rxin: actually still need this for union rdds

src/main/scala/shark/memstore2/CachePolicy.scala
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package shark.memstore2
+
+
+trait CachePolicy[K, V] {
+
+  protected var maxSize: Long = _
```
rxin: A style thing. Let's use _maxSize for the protected var, and use maxSize for the public getter (instead of getXXX).

src/main/scala/shark/memstore2/CachePolicy.scala
```diff
@@ -0,0 +1,53 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package shark.memstore2
+
+
+trait CachePolicy[K, V] {
```
rxin, line 34: In general add more comments to this class to explain this class and loadFunc / evictionFunc.

src/main/scala/shark/memstore2/CachePolicy.scala
rxin: and we can do this for hitRate, evictionCount, keysOfCachedEntries

src/main/scala/shark/SharkConfVars.scala
```diff
@@ -43,6 +43,23 @@ object SharkConfVars {
   // Default storage level for cached tables.
   val STORAGE_LEVEL = new ConfVar("shark.cache.storageLevel", "MEMORY_AND_DISK")
 
+  // True if a table's partitions should be evicted using a cache policy.
+  val SHOULD_USE_CACHE_POLICY = new ConfVar("shark.cache.partition.useCachePolicy", true)
```
rxin: (remove this config as well)

Let's just have the default policy be keep-all.

src/main/scala/shark/SharkConfVars.scala
```diff
@@ -43,6 +43,23 @@ object SharkConfVars {
...
+
+  // Class name of the default cache policy used to manage partition evictions for cached,
+  // Hive-partitioned tables.
+  val CACHE_POLICY = new ConfVar(
+    "shark.partition.cachePolicy.default", "shark.memstore2.LRUCachePolicy")
+
+  // Maximum size - in terms of the number of objects - of the cache specified by the
+  // "shark.cache.partition.cachePolicy" property above.
+  val MAX_PARTITION_CACHE_SIZE = new ConfVar("shark.cache.partition.cachePolicy.maxSize", "10")
```
rxin: We shouldn't put policy level configurations here because it makes it less pluggable (and this should be a table level setting really). Can we make this change?

In table property, the CachePolicy should also specify the parameters, .e.g

"shark.cache.partition.cachePolicy" = "shark.memstore2.LRACachePolicy(30)"

When we get the cache policy string, let's find the policy part (before parenthesis) and instantiate the policy, and leave the parts in parenthesis to the cache policy itself to parse.

rxin: also shark.cache.partition.cachePolicy is pretty verbose. How about "shark.cache.policy" (for the table property) ?

src/main/scala/shark/execution/RDDUtils.scala
rxin: Didn't Ram add some code to flatten union rdds? so there is always only one union rdd at the top, and no nested union rdds.

src/main/scala/shark/SharkConfVars.scala
```diff
@@ -43,6 +43,23 @@ object SharkConfVars {
   // Default storage level for cached tables.
   val STORAGE_LEVEL = new ConfVar("shark.cache.storageLevel", "MEMORY_AND_DISK")
 
+  // True if a table's partitions should be evicted using a cache policy.
+  val SHOULD_USE_CACHE_POLICY = new ConfVar("shark.cache.partition.useCachePolicy", true)
+
+  // Class name of the default cache policy used to manage partition evictions for cached,
+  // Hive-partitioned tables.
+  val CACHE_POLICY = new ConfVar(
+    "shark.partition.cachePolicy.default", "shark.memstore2.LRUCachePolicy")
```
rxin: and to make this consistent with the table property shortening, let's rename it to shark.cache.policy.default

src/main/scala/shark/SharkConfVars.scala
```diff
@@ -43,6 +43,23 @@ object SharkConfVars {
...
+
+  // Maximum size - in terms of the number of objects - of the cache specified by the
+  // "shark.cache.partition.cachePolicy" property above.
+  val MAX_PARTITION_CACHE_SIZE = new ConfVar("shark.cache.partition.cachePolicy.maxSize", "10")
+
+  // Whether the cache specified by "shark.cache.partition.cachePolicy" above should record
+  // statistics (e.g., hit rate).
+  val SHOULD_RECORD_PARTITION_CACHE_STATS = new ConfVar(
```
rxin: Remove this and make it always true.

src/main/scala/shark/SharkConfVars.scala
```diff
@@ -166,18 +184,18 @@ case class ConfVar(
   }
 
   def this(varname: String, defaultVal: Int) = {
-    this(varname, classOf[Int], null, defaultVal, 0, 0, false)
+    this(varname, classOf[Int], defaultVal.toString, defaultVal, 0, 0, false)
```
rxin, line 179: How come we need to set the default string value?

src/main/scala/shark/memstore2/MemoryMetadataManager.scala
```diff
@@ -52,63 +48,148 @@ class MemoryMetadataManager {
...
+    val lowerCaseTableName = tableName.toLowerCase
+
+    // Remove MemoryTable's entry from Shark metadata.
+    _keyToStats.remove(lowerCaseTableName)
+
+    val tableValue: Option[Table] = _keyToTable.remove(lowerCaseTableName)
+    return tableValue.flatMap(MemoryMetadataManager.unpersistRDDsInTable(_))
+  }
+
+  /** Find all keys that are strings. Used to drop tables after exiting. */
```
rxin: Make a TODO note to remove this for unified cache table view.

src/main/scala/shark/memstore2/MemoryMetadataManager.scala
```diff
@@ -52,63 +48,148 @@ class MemoryMetadataManager {
...
+  }
+
+  /**
+   * Return a representation of the partition key in the string format:
+   *     'col1=value1/col2=value2/...'
+   */
+  def makeHivePartitionKeyStr(
+      partitionColumns: Seq[String],
+      partitionColumnToValue: JavaMap[String, String]): String = {
+    var keyStr = ""
```
rxin: You can express this slightly more elegantly using

```
partitionColumns.map(col => "%s=%s/".format(col, partitionColumnToValue(col)).mkString("/"
```

src/main/scala/shark/memstore2/PartitionedMemoryTable.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  def containsPartition(partitionKey: String): Boolean = _keyToPartitions.contains(partitionKey)
+
+  def getPartition(partitionKey: String): Option[RDD[TablePartition]] = {
+    val rddValueOpt: Option[RDDValue] = _keyToPartitions.get(partitionKey)
+    if (rddValueOpt.isDefined && _cachePolicy.isDefined) _cachePolicy.get.notifyGet(partitionKey)
+    return rddValueOpt.map(_.rdd)
+  }
+
+  def putPartition(
```
rxin, line 74: Let's break this into two functions: insertIntoPartition and insertOverwritePartition

src/main/scala/shark/memstore2/PartitionedMemoryTable.scala
```diff
@@ -0,0 +1,141 @@
...
+  def keyToPartitions: collection.immutable.Map[String, RDD[TablePartition]] = {
+    return _keyToPartitions.mapValues(_.rdd).toMap
+  }
+
+  /**
+   * Computes the current storage level for this table. See comments in
+   * RDDUtils#getStorageLevelOfRDDs() for how Shark interprets the storage level of a sequence of
+   * RDDs.
+   */
+  override def getCurrentStorageLevel: StorageLevel = {
```
rxin: Again, remove this.

src/main/scala/shark/parse/SharkSemanticAnalyzer.scala
```diff
@@ -354,11 +340,94 @@ class SharkSemanticAnalyzer(conf: HiveConf) extends SemanticAnalyzer(conf) with
...
+    // If we detect that the CREATE TABLE is part of a CTAS, then this is set to the root node of
+    // the query command (i.e., the root node of the SELECT statement).
+    var queryStmtASTNode: Option[ASTNode] = None
+
+    // TODO(harvey): We might be able to reuse the QB passed into this method, as long as it was
+    //               created after the super.analyzeInternal() call. That QB and the createTableDesc
+    //               should have everything (e.g. isCTAS(), partCols). Note that the QB might not be
+    //               accessible from getParseContext(), since the SemanticAnalyzer#analyzeInternal()
+    //               doesn't set (this.qb = qb) for a non-CTAS.
+    var isRegularCreateTable = true
```
rxin, line 355: Add some comment explaining what a "regular" create table is.

src/main/scala/shark/execution/RDDUtils.scala
harveyfeng: Added a unionAndFlatten() in RDDUtils

src/main/scala/shark/memstore2/LRUCachePolicy.scala
harveyfeng: Replaced Guava Cache with a general LinkedHashMap, plus some bookkeeping of cache stats.

src/main/scala/shark/memstore2/LRUCachePolicy.scala
harveyfeng: Added a size and FIFO-based policy

src/main/scala/shark/SharkConfVars.scala
harveyfeng: Switched to "shark.cache.policy" and added some parsing helpers in CachePolicy.

The policy table property configs are here since I wasn't sure where to comment about them.
It seems like SharkConfVars only deals with session-wide Conf properties. Would it be alright to add a utility that only helps with table properties, something like a "SharkTblPropertyVars" Object? It'd double as a place to document those variables. Probably should be a new JIRA issue/future PR.

src/main/scala/shark/SharkConfVars.scala
harveyfeng, line 179: Part of being confused about where to put variables for conf vs tblproperties (a map)...

harveyfeng, line 179: A string value makes it easier to do getOrElse() for the maps

src/main/scala/shark/memstore2/CachePolicy.scala
```diff
@@ -0,0 +1,110 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.collection.JavaConversions._
+
+import org.apache.spark.rdd.RDD
+
+/**
+ * A cache that never evicts entries.
+ */
+class CacheAllPolicy[K, V] extends CachePolicy[K, V] {
```
rxin: Move this after CachePolicy is defined.

src/main/scala/shark/memstore2/LinkedMapBasedPolicy.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  override def initializeInternals(loadFunc: (K => V), evictionFunc: (K, V) => Unit) {
+    super.initializeInternals(loadFunc, evictionFunc)
+    _cache = new LinkedMapCache()
+  }
+
+}
+
+
+sealed abstract class LinkedMapBasedPolicy[K, V] extends CachePolicy[K, V] {
```
rxin: This file name is kind of weird right now - why don't we just move all these classes into the CachePolicy.scala file?

src/main/scala/shark/execution/RDDUtils.scala
```diff
@@ -34,22 +34,72 @@ import shark.SharkEnv
...
+    val unionedRdd = otherRdd match {
+      case unionRdd: UnionRDD[_] => {
+        new UnionRDD(rdd.context, (unionRdd.rdds :+ rdd))
+      }
+      case _ => rdd.union(otherRdd)
+    }
+    return unionedRdd
+  }
+
+  def getStorageLevelOfRDD(rdd: RDD[_]): StorageLevel = {
```
rxin: I suspect we don't need this anymore. If we need it for testing, perhaps move it into a testing class.

src/main/scala/shark/memstore2/CachePolicy.scala
```diff
@@ -0,0 +1,110 @@
...
+ */
+trait CachePolicy[K, V] {
+
+  protected var _loadFunc: (K => V) = _
+
+  protected var _evictionFunc: (K, V) => Unit = _
+
+  protected var _maxSize: Int = -1
+
+  def initializeWithUserSpecs(args: Array[String], fallbackMaxSize: Int) {
```
rxin: Let's simplify this by moving this into the constructor. Then we don't need the extra step in init nor the assert in every function.

rxin:
Thanks, Harvey, for this giant change!

PR #185, opened 2013-10-08
Minimize Configuration broadcast/serialization overhead for HadoopRDDs/Hive-partition scans.

harveyfeng:
Previously, there was one broadcasted Configuration for each HadoopRDD created for a Hive partition that's stored on disk - e.g., for a table scan over 1000 partitions, there are 1000 Configuration broadcasts.
This change uses the HadoopFileRDD, so that there is one broadcasted Configuration shared across multiple HadoopRDDs created for the table scan.

Also, moved row deserialization from processPartition() to preprocessRDD(). This makes it more consistent with partitioned table reads/processing, and allows us to annotate HiveConf as transient, which should reduce Spark task sizes.

### DIFF
--- a/src/main/scala/shark/execution/TableScanOperator.scala
+++ b/src/main/scala/shark/execution/TableScanOperator.scala
@@ -19,6 +19,7 @@ package shark.execution
 
 import java.util.{ArrayList, Arrays}
 import scala.reflect.BeanProperty
+import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.mapred.{FileInputFormat, InputFormat, JobConf}
 import org.apache.hadoop.hive.conf.HiveConf
 import org.apache.hadoop.hive.metastore.api.Constants.META_TABLE_PARTITION_COLUMNS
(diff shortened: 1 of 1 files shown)


src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -214,15 +217,53 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
    * Create a RDD representing the table (with or without partitions).
    */
   override def preprocessRdd(rdd: RDD[_]): RDD[_] = {
+    // copyTableJobPropertiesToConf and pushFilters both take in JobConfs.
+    val localJobConf = new JobConf(localHconf)
```
rxin: Can we move the job conf setting into its own function?

rxin:
Hi Harvey - this looks great. 

I had only one comment, and please make sure this builds with your latest change to Spark.

harveyfeng:
TableScanOperator updated, PTAL

harveyfeng:
Going to reopen a fresh PR tomorrow, once the Spark API changes are reflected.

PR #187, opened 2013-10-15
Revert to using RDD#collect() for map joins,

harveyfeng:
Since Spark can send Task results through the block manager when the Akka frame size is exceeded.

Notes:
- Includes patch from https://github.com/amplab/shark/pull/188, which fixes a group-by bug.
- Reverts Shark PR from https://github.com/amplab/shark/pull/98
- Relevant commit on the Spark side: https://github.com/apache/incubator-spark/commit/c75eb14fe52f6789430983471974e5ddf73aacbf


rxin:
Thanks. I am going to merge this first so it can compile.

PR #188, opened 2013-10-15
Add a NULL check in ReduceKeyReduceSide#equals().

harveyfeng:
This bug was revealed by the AppendOnlyMap.


harveyfeng:
Combining this with https://github.com/amplab/shark/pull/188 so that it can compile....

PR #189, opened 2013-10-16
Get Shark to compile with latest Spark master.

harveyfeng:
This seems to be the only change, after relaxing access constraints on TaskContext's stageId here: https://github.com/apache/incubator-spark/pull/62

Also included removing Kryo registration, since Shark uses its ShuffleSerializer for serializing objects during shuffles. This also fixes an earlier issue with mapjoin serialization here: https://github.com/amplab/shark/pull/187


rxin:
Thanks.

PR #190, opened 2013-10-16
Minimize Configuration broadcast/serialization overhead for HadoopRDDs/Hive-partition scans.

harveyfeng:
Note: relies on https://github.com/amplab/shark/pull/189 

Previously, there was one broadcasted Configuration for each HadoopRDD created for a Hive partition that's stored on disk - e.g., for a table scan over 1000 partitions, there are 1000 Configuration broadcasts.
This change uses the HadoopRDD constructor added recently (https://github.com/apache/incubator-spark/commit/5a99e67894e3a7716a8a1eaa19ce17e7a8abf5b6), which allows for one broadcasted Configuration to be shared across multiple HadoopRDDs created for the table scan.

Also, moved HadoopRDD row deserialization from processPartition() to preprocessRDD() in TableScanOperator. This makes it more consistent with how partitioned tables  are processed, and allows HiveConf to be annotated as transient, which should reduce Spark task sizes.


src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -289,55 +335,127 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
...
+        }
+        columnTypes.append(columnInfo.getType().getTypeName())
+      }
+      val columnTypesString = columnTypes.toString
+      hiveConf.set(Constants.LIST_COLUMN_TYPES, columnTypesString)
+    }
+
+    // Push down predicate filters.
+    val filterExprNode = tableScanDesc.getFilterExpr()
+    if (filterExprNode == null) return
```
rxin: Can we rewrite this as 

```
if (filterExprNode != null) {
....
}
```

src/main/scala/shark/execution/TableScanOperator.scala
```diff
@@ -211,31 +223,62 @@ class TableScanOperator extends TopOperator[HiveTableScanOperator] with HiveTopO
   }
 
   /**
-   * Create a RDD representing the table (with or without partitions).
+   * Create a RDD for a table.
    */
   override def preprocessRdd(rdd: RDD[_]): RDD[_] = {
+    // Choose the minimum number of splits. If mapred.map.tasks is set, use that unless
```
rxin, line 229: Let's change execute to not call super.execute (so we can skip the extra RDD creation in processPartition)

src/main/scala/shark/execution/TableScanOperator.scala
harveyfeng: done

src/main/scala/shark/execution/TableScanOperator.scala
harveyfeng, line 229: done

harveyfeng, line 229: also renamed preprocessRdd() to makeRDDFromHadoop(), to make it a bit more clear.

rxin:
Thanks, Harvey.

PR #191, opened 2013-10-17
HiveConfSerializer code cleanup

sundeepn:
This code is not used and the intended optimization is covered else where. This code/tests cause issues with hadoop 0.23.9. Cleaning up.


rxin:
Jenkins, test this please.

rxin:
Thanks, Sundeep. The test failed for an pre-existing problem. I am merging this now.

PR #195, opened 2013-10-18
Enable query cancellation in local CLI.

rxin:


src/main/scala/shark/SharkCliDriver.scala
```diff
@@ -200,7 +211,7 @@ object SharkCliDriver {
       }
       if (line.trim().endsWith(";") && !line.trim().endsWith("\\;")) {
         line = prefix + line
-        ret = cli.processLine(line)
+        ret = cli.processLine(line, true)
```
harveyfeng, line 215: Format to:
`cli.processLine(line, allowInterrupting = true)`

rxin, line 215: That wouldn't work for Java source code, only Scala.

src/main/scala/shark/SharkCliDriver.scala
```diff
@@ -25,33 +25,44 @@ import java.io.PrintStream
...
-  var prompt2 = "     " // when ';' is not yet seen.
+  private var prompt  = "shark"
+  private var prompt2 = "     " // when ';' is not yet seen.
+
+  installSignalHandler()
+
+  /**
+   * Install a interrupt callback to cancel all Spark jobs. In Hive CliDriver's processLine, it
+   * installs a signal handler for ctrl + c to invoke callbacks defined in HiveInterruptUtils.
+   */
```
harveyfeng, line 59: Would it make more sense to rephrase this comment to:
Install an interrupt callback to cancel all Spark jobs. In Hive's CliDriver#processLine(), a signal handler will invoke this registered callback if a Ctrl+C signal is detected while a command is being processed by the current thread.

PR #197, opened 2013-10-23
Catch exceptions thrown by threads launched in ScriptOperator and get Shark to compile with latest Spark master.

harveyfeng:
Two additions in this PR, but both are relatively small and needed to get Jenkins passing.
- Exceptions thrown by threads launched in ScriptOperator are now caught and rethrown by on-complete-callbacks registered with the TaskContext used to compute the corresponding RDD.
- SparkHadoopWriter#createJobID() access was changed here: https://github.com/apache/incubator-spark/pull/88, but the fix is pretty small - just copied createJobID() and made it a nested function in FileSinkOperator...

### DIFF
--- a/src/main/scala/shark/execution/ScriptOperator.scala
+++ b/src/main/scala/shark/execution/ScriptOperator.scala
@@ -20,9 +20,11 @@ package shark.execution
 import java.io.{File, InputStream, IOException}
 import java.lang.Thread.UncaughtExceptionHandler
 import java.util.{Arrays, Properties}
+
 import scala.collection.JavaConversions._
 import scala.io.Source
 import scala.reflect.BeanProperty
+
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.hive.conf.HiveConf
 import org.apache.hadoop.hive.ql.exec.{ScriptOperator => HiveScriptOperator}
@@ -32,9 +34,11 @@ import org.apache.hadoop.hive.ql.plan.ScriptDesc
 import org.apache.hadoop.hive.serde2.{Serializer, Deserializer}
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector
 import org.apache.hadoop.io.{BytesWritable, Writable}
+
 import org.apache.spark.{OneToOneDependency, SparkEnv, SparkFiles}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.TaskContext
+
 import shark.execution.serialization.OperatorSerializationWrapper
 import shark.LogHelper
 
@@ -119,7 +123,6 @@ class ScriptOperator extends UnaryOperator[HiveScriptOperator] {
           SparkEnv.set(sparkEnv)
           val recordWriter = inRecordWriterClass.newInstance
           recordWriter.initialize(proc.getOutputStream, op.localHconf)
-
           for (elem <- iter) {
             recordWriter.write(elem)
           }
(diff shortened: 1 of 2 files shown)


src/main/scala/shark/execution/ScriptOperator.scala
```diff
@@ -228,6 +238,45 @@ class ScriptOperator extends UnaryOperator[HiveScriptOperator] {
...
   /**
+   * A general exception handler to attach to child threads used to feed input rows and forward
+   * errors to the parent thread during ScriptOperator#execute().
+   * If partial query consumption is not allowed (see HiveConf.Confvars.ALLOWPARTIALCONSUMP), then
+   * exceptions from child threads are caught by the handler and re-thrown by the parent thread
+   * through an on-task-completion callback registered with the Spark TaskContext. The task will be
+   * marked "failed" and the exception will be propagated to the master/CLI.
+   */
+  class ScriptExceptionHandler(
+      serializedScriptOp: OperatorSerializationWrapper[ScriptOperator], context: TaskContext)
```
rxin: Let's just pass allowPartialConsumption here instead of the entire operator.

src/main/scala/shark/execution/ScriptOperator.scala
```diff
@@ -228,6 +238,45 @@ class ScriptOperator extends UnaryOperator[HiveScriptOperator] {
...
+          if (serializedScriptOp.allowPartialConsumption) {
+            logWarning("Error while executing script. Ignoring %s"
+              .format(ioe.getMessage))
+          } else {
+            val onCompleteCallback = () => {
+              logWarning("Error during script execution. Set %s=true to ignore thrown IOExceptions."
+                .format(HiveConf.ConfVars.ALLOWPARTIALCONSUMP.toString))
+              throw ioe
+            }
+            context.addOnCompleteCallback(onCompleteCallback)
```
rxin: We need to synchronize on context here because addOnCompleteCallback is not thread safe.

src/main/scala/shark/execution/ScriptOperator.scala
```diff
@@ -228,6 +238,45 @@ class ScriptOperator extends UnaryOperator[HiveScriptOperator] {
...
+          }
+        }
+        case _ => {
+          // Throw any other Exceptions or Errors.
+          val onCompleteCallback = () => throw throwable
+          context.addOnCompleteCallback(onCompleteCallback)
```
rxin: sychronized on context here too.

rxin: I wonder if there is a racing condition here - what if the script program shuts down, and we catch the exception, but the throwable callback is registered after the consumer thread (i.e. the main thread) is done?

When the script program shuts down, does the main thread get any error?

harveyfeng: If one of the child threads registers the callback after the entire task finishes and TaskContext#executeOnCompleteCallbacks() is run, then the main thread never executes that callback, and everything finishes silently.

Manually tested that part by adding a `Thread.sleep(10000)`...

src/main/scala/shark/execution/ScriptOperator.scala
harveyfeng: done

src/main/scala/shark/execution/ScriptOperator.scala
harveyfeng: done.

rxin:
Thanks!

PR #200, opened 2013-10-26
Increased the number of args for RDDTable#apply() to 22, the max Scala tuple length.

harveyfeng:
RDDTableFunctions and RDDTable have their own files now, too.


rxin:
Thanks!

PR #203, opened 2013-10-30
Add a TableReader trait and implementations for scanning tables stored in Hadoop, Tachyon, and Spark's block manager.

harveyfeng:
Mostly refactoring, the TableReader implementations decouple main table scan execution paths from TableScanOperator. However, pruning RDDs is still dependent on operator-specific executions, so that hasn't been ported.

The main benefit is that TableReaders can be reused for other scenarios, such as SharkLoadTask or TableToRDD execution.

### DIFF
--- a/src/main/scala/shark/execution/TableReader.scala
+++ b/src/main/scala/shark/execution/TableReader.scala
@@ -55,6 +55,7 @@ trait TableReader extends LogHelper{
 
 }
 
+/** Helper class for scanning tables stored in Tachyon. */
 class TachyonTableReader(@transient _tableDesc: TableDesc) extends TableReader {
 
   // Split from 'databaseName.tableName'
@@ -89,6 +90,7 @@ class TachyonTableReader(@transient _tableDesc: TableDesc) extends TableReader {
 
 }
 
+/** Helper class for scanning tables stored in Spark's block manager */
 class HeapTableReader(@transient _tableDesc: TableDesc) extends TableReader {
 
   // Split from 'databaseName.tableName'
@@ -96,6 +98,7 @@ class HeapTableReader(@transient _tableDesc: TableDesc) extends TableReader {
   private val _databaseName = _tableNameSplit(0)
   private val _tableName = _tableNameSplit(1)
 
+  /** Fetches the RDD for `_tableName` from the Shark metastore. */
   override def makeRDDForTable(hiveTable: HiveTable): RDD[_] = {
     logInfo("Loading table %s.%s from Spark block manager".format(_databaseName, _tableName))
     val tableOpt = SharkEnv.memoryMetadataManager.getMemoryTable(_databaseName, _tableName)
@@ -109,7 +112,6 @@ class HeapTableReader(@transient _tableDesc: TableDesc) extends TableReader {
    * Fetch an RDD from the Shark metastore using each partition key given, and return a union of all
    * the fetched RDDs.
    *
-   * @param tableKey Name of the partitioned table.
    * @param partitions A collection of Hive-partition metadata, such as partition columns and
    *     partition key specifications.
    */
@@ -162,6 +164,10 @@ class HeapTableReader(@transient _tableDesc: TableDesc) extends TableReader {
     }
   }
 
+  /**
+   * Thrown if the table identified by the (_databaseName, _tableName) pair cannot be found in
+   * the Shark metastore.
+   */
   private def throwMissingTableException() {
     logError("""|Table %s.%s not found in block manager.
                 |Are you trying to access a cached table from a Shark session other than the one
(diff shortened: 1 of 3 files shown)


src/main/scala/shark/execution/TableReader.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  override def makeRDDForTable(hiveTable: HiveTable): RDD[_] = {
+    // Table is in Tachyon.
+    val tableKey = SharkEnv.makeTachyonTableKey(_databaseName, _tableName)
+    if (!SharkEnv.tachyonUtil.tableExists(tableKey)) {
+      throw new TachyonException("Table " + tableKey + " does not exist in Tachyon")
+    }
+    logInfo("Loading table " + tableKey + " from Tachyon.")
+
+    var indexToStats: collection.Map[Int, TablePartitionStats] =
```
rxin: Should this one be null? Or better, do

```
if (!SharkEnv.memoryMetadataManager.getStats(_databaseName, _tableName).isDefined) {
  ...
}
```

Right now it sounds as indexToStats is to be used in this function.

rxin: Let's also take this chance to add some comment to explain why we are doing this (if the stats are stored in tachyon, and then shark is restarted, mem meta data mgr wouldn't have the stats data).

src/main/scala/shark/execution/TableReader.scala
```diff
@@ -0,0 +1,182 @@
...
+      indexToStats = JavaSerializer.deserialize[collection.Map[Int, TablePartitionStats]](
+        statsByteBuffer.array())
+      logInfo("Loading table " + tableKey + " stats from Tachyon.")
+      SharkEnv.memoryMetadataManager.putStats(_databaseName, _tableName, indexToStats)
+    }
+    SharkEnv.tachyonUtil.createRDD(tableKey)
+  }
+
+  override def makeRDDForPartitionedTable(partitions: Seq[Partition]): RDD[_] = {
+    throw new UnsupportedOperationException("Partitioned tables are not yet supported for Tachyon.")
```
rxin, line 85: We should work with @haoyuan to get this working for Tachyon.

src/main/scala/shark/execution/TableReader.scala
harveyfeng: Yeah, I think it's cleaner to use Option#isDefined/isEmpty...this type of `null` pattern appears in other parts too, such as for TachyonWriter and TablePartitionBuilder construction in MemoryStoreOperator. I'll fix that in another patch that will try to decouple RDD[TablePartition] construction and materialization from MemoryStoreSinkOperator.

src/main/scala/shark/execution/TableReader.scala
```diff
@@ -0,0 +1,182 @@
...
+  // Split from 'databaseName.tableName'
+  private val _tableNameSplit = _tableDesc.getTableName.split('.')
+  private val _databaseName = _tableNameSplit(0)
+  private val _tableName = _tableNameSplit(1)
+
+  /** Fetches the RDD for `_tableName` from the Shark metastore. */
+  override def makeRDDForTable(hiveTable: HiveTable): RDD[_] = {
+    logInfo("Loading table %s.%s from Spark block manager".format(_databaseName, _tableName))
+    val tableOpt = SharkEnv.memoryMetadataManager.getMemoryTable(_databaseName, _tableName)
+    if (tableOpt.isEmpty) throwMissingTableException()
```
rxin: Wrap the body in { }

```
if (tableOpt.isEmpty) {
  throwMissingTableException()
}
```

also do it for the other place that calls throwMissingTableException

src/main/scala/shark/execution/TableReader.scala
```diff
@@ -0,0 +1,182 @@
...
+ */
+
+package shark.execution
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.PathFilter
+import org.apache.hadoop.hive.conf.HiveConf
+import org.apache.hadoop.hive.metastore.api.Constants.META_TABLE_PARTITION_COLUMNS
+import org.apache.hadoop.hive.ql.exec.Utilities
+import org.apache.hadoop.hive.ql.metadata.{Partition, Table => HiveTable}
```
rxin: Also rename Partition to HivePartition otherwise it is confusing in the code

src/main/scala/shark/execution/TableReader.scala
```diff
@@ -0,0 +1,182 @@
...
+import org.apache.hadoop.hive.ql.plan.{PartitionDesc, TableDesc}
+
+import org.apache.spark.rdd.{EmptyRDD, RDD, UnionRDD}
+import org.apache.spark.broadcast.Broadcast
+import org.apache.spark.SerializableWritable
+
+import shark.api.QueryExecutionException
+import shark.execution.optimization.ColumnPruner
+import shark.execution.serialization.JavaSerializer
+import shark.{LogHelper, SharkConfVars, SharkEnv}
```
rxin: Move this import before shark.api

harveyfeng:
Thanks for the review! All comments addressed.

rxin:
Thanks. Merged this!

PR #169, opened 2013-09-19
scalastyle checks for sbt. All warnings for now. Execute using 'sbt scal...

junkumar:
...astyle'

Reynold, take a look and see if these simple checks for you. They include 100 columns per line, braces, whitespace etc..

PR #205, opened 2013-11-02
Merging Nandu's style cleanup pull request.

rxin:
I did a manual merge to bring #169 into master.


src/main/scala/shark/util/MurmurHash3_x86_128.scala
```diff
@@ -179,16 +196,16 @@ object MurmurHash3_x86_128 {
...
       case 1 => data(index) << 24
       case 0 => 0
       case _ => data(index) << 24 | 
-      			(data(index + 1) & 0xFF) << 16 |
-      			(data(index + 2) & 0xFF) << 8 |
-      			(data(index + 3) & 0xFF)
+                (data(index + 1) & 0xFF) << 16 |
+                (data(index + 2) & 0xFF) << 8 |
+                (data(index + 3) & 0xFF)
     }
```
harveyfeng, line 209: Some of the `case` statement expressions span multiple lines, but it's not necessary to wrap them in curly braces, since they're fairly clear to follow?
The Scala style guide does mention that curly braces should be used if there's any muti-lined expression, but it's a bit ambiguous for cases where there's a single expression that requires line wrapping...
Most of the Twitter code (e.g, algebird) omits curly braces for similar code structures like this.

src/main/scala/shark/util/MurmurHash3_x86_128.scala
```diff
@@ -179,16 +196,16 @@ object MurmurHash3_x86_128 {
   @inline final def getInt(data: Array[Byte], index: Int, rem: Int): Int = {
     rem  match {
```
harveyfeng: There are 2 spaces here

src/main/scala/shark/util/BloomFilter.scala
```diff
@@ -98,7 +115,7 @@ class BloomFilter(numBitsPerElement: Double, expectedSize: Int, numHashes: Int)
    * @param data is the bytes to be hashed.
    * @param length is the length of the buffer to examine.
    * @return true with some false positive probability and false if the
-   * 			bytes is not contained in the bloom filter.
+   * bytes is not contained in the bloom filter.
```
harveyfeng: Could this comment be inlined with respect to `@return`, or indented 4 spaces?

src/main/scala/shark/execution/MapSplitPruning.scala
```diff
@@ -135,19 +138,19 @@ object MapSplitPruning {
...
-      else if (right.isInstanceOf[ExprNodeColumnEvaluator])
+      } else if (right.isInstanceOf[ExprNodeColumnEvaluator]) {
         right.asInstanceOf[ExprNodeColumnEvaluator]
-      else null
+      } else null
 
     // Try to get the constant value.
     val constEval: ExprNodeConstantEvaluator =
-      if (left.isInstanceOf[ExprNodeConstantEvaluator])
+      if (left.isInstanceOf[ExprNodeConstantEvaluator]) {
```
harveyfeng, line 151: Maybe surround the `else null`here and on L141 with curly braces, to be consistent? Though the Scala style guide mentions that omitting braces is okay if all the branches are single line expressions (http://docs.scala-lang.org/style/control-structures.html)
It might also be nicer to put the `if (...)` on the same line as the `constEval` declaration, to save a line/indents (applies to the block at L141 too).

src/main/scala/shark/execution/serialization/ShuffleSerializer.scala
```diff
@@ -22,7 +22,8 @@ import java.nio.ByteBuffer
 
 import org.apache.hadoop.io.BytesWritable
 
-import org.apache.spark.serializer.{DeserializationStream, Serializer, SerializerInstance, SerializationStream}
+import org.apache.spark.serializer.DeserializationStream
+import org.apache.spark.serializer.{Serializer, SerializerInstance, SerializationStream}
```
harveyfeng: SerializationStream should be first

rxin:
Ok all comments addressed (except the last one which could go either way)

PR #209, opened 2013-11-04
fix byte buffer overflow caused by wrong size info

xiliu82:
Do I need to send double pull requests to both master and branch?


rxin:
I usually cherry pick the commits into branch-0.8. Thanks though. This makes it easier.

rxin:
Jenkins, test this please.

rxin:
Thanks for doing this!

PR #210, opened 2013-11-05
FIXED SHARK-205: Map partition pruning for string column gets NPE on empty partitions.

rxin:


rxin:
@sundeepn

@harsha2010

harsha2010:
Hi Reynold

Thanks, this looks good.

sundeepn:
Looks good to me. Thanks for the fix.

PR #211, opened 2013-11-05
Improve the RDD <-> Table conversion support

alig:


src/main/resources/tablerdd/table_rdd_generators.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/python
+
+import sys
```
rxin: Maybe add some comments here about this file to explain what it is used for, and how it should be run?

rxin:
Jenkins, add to whitelist.

src/main/scala/shark/SharkContext.scala
```diff
@@ -104,6 +105,143 @@ class SharkContext(
...
+   * a SELECT statement.
+   */
+  def sqlRdd[T](cmd: String): RDD[Tuple1[T]] =
+    new TableRDD1[T](sqlRowRdd(cmd))
+
+  private type M[T] = ClassManifest[T]
+  private def m[T](implicit m : ClassManifest[T]) = classManifest[T](m)
+
+  def sqlRdd[T1: M,T2: M](cmd: String):
+  RDD[Tuple2[T1,T2]] =
```
rxin: Can we make the Python script generate code that indents this way?

```
def sqlRdd[T1: M, T2: M](cmd: String):
  RDD[Tuple2[T1, T2]] = {
  new TableRDD2[T1, T2](sqlRowRdd(cmd),
    Seq(m[T1], m[T2]))
}
```

(and add a whitespace after comma)

src/main/scala/shark/SharkContext.scala
```diff
@@ -91,7 +92,7 @@ class SharkContext(
    * Execute a SQL command and return the results as a TableRDD. The SQL command must be
    * a SELECT statement.
    */
-  def sql2rdd(cmd: String): TableRDD = {
+  def sqlRowRdd(cmd: String): RowRDD = {
```
rxin: This is API breaking. We need to discuss this.

JoshRosen:
If you want to use Python to generate code, check out [cog](http://nedbatchelder.com/code/cog/).  It would let you embed Python code generation statements inside of comments in the .Scala file, so you don't have to copy-paste the generated code to update it.  I think it might also have some functions to help with indentation.  In undergrad, I used it to generate C code.

src/main/resources/tablerdd/table_rdd_generators.py
```diff
@@ -0,0 +1,58 @@
...
+            res     += sep
+            oneLine += sep
+            if len(oneLine) > newlineAfter:
+                res += "\n" + " "*indent
+                oneLine = ""
+    return res
+
+### The SparkContext declaration
+for x in range(2,23):
+    p.write("class TableTupleRDD" + str(x) + "[")
```
JoshRosen: For readability, I'd consider building these from [template strings](http://docs.python.org/2/library/string.html#template-strings).

src/main/scala/shark/api/TableRDD.scala
```diff
@@ -17,64 +17,614 @@
...
+  def getSchema = prev.schema
 
-import org.apache.spark.{Partition, TaskContext}
-import org.apache.spark.rdd.RDD
+  override def getPartitions = prev.getPartitions
+
+  override def compute(split: Partition, context: TaskContext):
+  Iterator[Seq[Any]] = {
+    prev.compute(split, context).map( row =>
+      (0 until prev.schema.size).map(i => row.getPrimitive(i)) )
```
harveyfeng: If the Range is moved to the outer block, then it's ~~computed~~ created once per partition rather than once per row.
(though maybe compiler magic alleviates it, or a Range object is small enough that it doesn't matter...)

Would it also be cleaner - for the implementation and API - if the row object had a:

```
  def toSeq = {
    oi.getAllStructFieldRefs.map { structField =>
      val primitiveData = oi.getStructFieldData(rawdata, structField)
      structField.getFieldObjectInspector.asInstanceOf[PrimitiveObjectInspector].
        getPrimitiveJavaObject(primitiveData)
    }
  }
```

and maybe a `toTuple`?

src/main/scala/shark/SharkContext.scala
```diff
@@ -104,6 +105,187 @@ class SharkContext(
...
+
+  /**
+   * Execute a SQL command and return the results as a RDD of Tuple. The SQL command must be
+   * a SELECT statement.
+   */
+  def sqlRdd[T](cmd: String): RDD[Tuple1[T]] =
+    new TableRDD1[T](sqlRowRdd(cmd))
+
+  private type M[T] = ClassManifest[T]
+  private def m[T](implicit m : ClassManifest[T]) = classManifest[T](m)
```
harveyfeng: Move these before function declarations

src/main/scala/shark/SharkContext.scala
```diff
@@ -104,6 +105,187 @@ class SharkContext(
...
+
+  def sqlRdd[T1: M, T2: M](cmd: String):
```
harveyfeng, line 125: Might want to add a comment somewhere before all the sqlRdd() methods and the TableRDD classes stating that they were generated from a script,  just to go with the general convention that it's okay for generated code to be messy...

src/main/scala/shark/api/TableRDD.scala
harveyfeng: Looking at this again, it also seems like StructObjectInspector#getAllStructFieldRefs() returns a list, so Row#getPrimitive(i) might be slower than the toSeq above...

AndreSchumacher:
I just tried it. The direction table -> RDD works great. Looking at RDDTableFunctions it should also be possible to store an RDD as a table but is there an implicit conversion missing, maybe in SharkContext?

harveyfeng:
There is an `RDDTable(someRdd).saveAsTable(tableName, columnNames)` that works for`someRdd: RDD[Tuple]`. Extending that API to support saving RDDs of sequences or rows should be doable using some implicits in SharkContext.
An implicit conversion in the SharkContext to get an RDD->RDDTableFunctions for any `saveAsTable()` does seem more intuitive, though.

src/main/scala/shark/api/TableRDD.scala
harveyfeng: nvm, the structs best use array-backed lists for those fields

AndreSchumacher:
Thanks, @harveyfeng . Turns out I was "holding it wrong" and assumed one should use sqlRowRDD for loading the RDD. This here works:

``` scala
import shark.api.RDDTable
val tmp = sc.sqlRdd[Integer,String]("select * from src")
tmp.collect()
RDDTable(tmp).saveAsTable("tmp_table", List("key", "value"))
sc.runSql("select * from tmp_table")
```

I was just wondering: is there any reason to expose RowRDD to users?

alig:
Well, the crux right now is that the Tuples only give you up to 22. Seq would be an alternative, but we should maybe make it lazy.

mateiz:
Is there an implicit conversion from RDD to RDDTable yet? We should add that so you can also do `rdd.saveAsTable(...)` instead of `RDDTable(rdd).saveAsTable(...)`.

AndreSchumacher:
+1 on the implicit conversion. Also I noticed that saved tables are by default cached (which makes sense) but there should be also a way to persist them to disk. Otherwise when one stores them inside shark-shell's scala shell they will show up inside shark's (shark) shell but you get an exception when running any operations on them.

BTW: I'd also volunteer to help out on this if there is need. I am very interested in this from a user point-of-view since I'm trying to import RDD's with data that Hive doesn't support import for (Matt's Parquet/Avro genomics input formats). That's a nice backdoor to get that into Shark.

rxin:
@AndreSchumacher please do chime in on this one :)

alig:
Good idea @mateiz and @AndreSchumacher, added implicits. I assume you have to do `import shark.api.RDDTableImplicits._` somewhere.

AndreSchumacher:
So I tried to find a simpler way to convert the MemoryTable the RDD is turned into to a "regular" Hive-like table whose data is persisted to disk. Please someone ( @rxin or @alig ?) correct me if I am wrong but it seems there is no easy way to do that without generating all the Hive storage-related metadata (FileSinkDesc, possibly parts of TableDesc) from scratch.

So as a workaround (see PR to Ali's clone's master branch) I propose to use SharkContext's runsql for all but the default (MEMORY_ONLY) storage level. It won't work if SharkContext is not available though and one gets an exception. Maybe that could be treated more gracefully maybe.

rxin:
@AndreSchumacher can you elaborate on runSql? runSql collects the output to the driver node.

AndreSchumacher:
Sorry for the confusing comments. In that workaround I run a "CREATE TABLE Y AS SELECT \* FROM X" inside the runSql, where X is the memory-backed (temporary) table that was generated from the RDD and Y is the table that is then also backed by HDFS, if the user specifies anything else than MEMORY_ONLY. So there isn't any data coming over to the driver node, it's just the DDL task essentially.

rxin:
Take a look at #213 - I think this is solved there already.

AndreSchumacher:
Thanks! That indeed looks very relevant and may fix the problem. I'll try it out and report back.

src/main/scala/shark/api/RDDTable.scala
```diff
@@ -19,12 +19,98 @@ package shark.api
...
+  RDDTableFunctions = RDDTable(rdd)
+}
 
 object RDDTable {
 
   private type M[T] = ClassManifest[T]
   private def m[T](implicit m : ClassManifest[T]) = classManifest[T](m)
 
+  /*
+   * NB: These functions are auto-generated using resources/tablerdd/table_rdd_generators.py
```
rxin: Are there more functions here? Otherwise it doesn't match the comment.

src/main/scala/shark/api/TableRDD.scala
```diff
@@ -20,14 +20,13 @@ package shark.api
 import java.util.{List => JList}
 
 import org.apache.hadoop.hive.metastore.api.FieldSchema
-import org.apache.hadoop.hive.serde2.objectinspector.{ObjectInspector, StructObjectInspector}
+import org.apache.hadoop.hive.serde2.objectinspector.{PrimitiveObjectInspector, ObjectInspector, StructObjectInspector}
```
rxin: Do you mind undoing this file?

src/main/resources/tablerdd/table_rdd_generators.py
```diff
@@ -0,0 +1,75 @@
...
+
+    tableClass = Template(
+"""
+class TableRDD$num[$list](prev: RowRDD,
+                       mans: Seq[ClassManifest[_]])
+  extends RDD[Tuple$num[$list]](prev) {
+  def schema = prev.schema
+
+  val tableCols = schema.size
+  if (tableCols != $num) throw new IllegalArgumentException("Table only has " + tableCols + " columns, expecting $num")
```
rxin: Change this to 

```
require(tableCols == $num, "Table only has " + tableCols + " columns, expecting $num")
```

Also, do you actually want tableCols to be publicly accessible?

src/main/scala/shark/api/TableRDDGenerated.scala
```diff
@@ -0,0 +1,630 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package shark.api
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.{TaskContext, Partition}
+
+class TableSeqRDD(prev: TableRDD)
```
rxin, line 27: Maybe take this out, and then this file is entirely generated. Can we add some comment saying this file is generated?

src/main/scala/shark/api/TableRDDGenerated.scala
```diff
@@ -0,0 +1,630 @@
...
+  extends RDD[Seq[Any]](prev) {
+
+  def getSchema = prev.schema
+
+  override def getPartitions = prev.getPartitions
+
+  override def compute(split: Partition, context: TaskContext):
+  Iterator[Seq[Any]] = {
```
rxin:
```
override def compute(split: Partition, context: TaskContext):Iterator[Seq[Any]] = {
```

(i.e. put it on the same line)

src/main/scala/shark/SharkContext.scala
```diff
@@ -104,6 +109,154 @@ class SharkContext(
...
+   * NB: These are auto-generated using resources/tablerdd/table_rdd_generators.py
+   */
+  def sqlSeqRdd(cmd: String): RDD[Seq[Any]] =
+    new TableSeqRDD(sql2rdd(cmd))
+
+  /**
+   * Execute a SQL command and return the results as a RDD of Tuple. The SQL command must be
+   * a SELECT statement.
+   */
+  def sqlRdd[T](cmd: String): RDD[Tuple1[T]] =
```
rxin: Change the code generator to put { } around the function body, since many of these generated functions become fairly long.

alig:
I have now changed the `saveAsTable()` functionality to mainly work with `Seq[_]` rather than Tuples. The reason for this is that in some settings you have more than 22 columns and Tuples would be insufficient. The implicit type conversions from RDD's can however convert `RDD[Product]` to `RDD[Seq[_]]` so users should be able to to save tables that are `RDD[Product]` (i.e. Tuples) and `RDD[Seq[_]]`.

Furthermore, I believe I addressed all the PR comments. I cleaned up the generated scripts and separated them into 3 different files. Two of them now generate their corresponding file in its entirety (the third one generates part of SharkContext). The filenames now match the files they generate and there is an auto-generate comment inside them.

alig:
Jenkins, please test this!

rxin:
Thanks - merging this now!

PR #213, opened 2013-11-10
New semantics for caching tables

harveyfeng:
Enables Shark to maintain a "unified view" of cached tables, in that its in-memory, columnar-backed store is synchronized with the backing disk store (the table's data directory in the Hive warehouse). Both the cached and on-disk copies are modified for any table updates (e.g., specified by `LOAD`, `INSERT`, or `CREATE TABLE AS SELECT`) - the behavior is similar to a write-through cache.

In terms of usability,
- No longer required to create new tables in order to cache existing ones.
  <p> Instead of `CREATE TABLE src_cached AS SELECT * FROM src`, tables can be cached using `CACHE src`. Uncaching is just an `UNCACHE src`
- Since unified views are backed by a disk copy, updates on cached tables persist across Shark sessions.
  <p> Previously, the `src_cached` table and any updates to it were only accessible from the Shark session that it was created in. Accessing it from a different session would cause a thrown runtime exception (something like a `QueryExecutionException: Table src not found in block manager`).
  By default, all cached tables are unified. This can be switched off by setting `"shark.cache" = "false"` in the table properties, or calling `set shark.cache.unify.default = false` in the Hive CLI.

In terms of implementation, the main additions for handling commands targeting unified views are: 
- Maintenance of two SerDes
  <p> All unified views have `LazySimpleSerDe` as its SerDe property. `TableScanOperator` will use a `ColumnarSerDe` if it notices that the table scan is targeted at an in memory table.
- Different  query plans generated for `INSERT` and `CTAS`.
  <p> Previously, the `MemoryStoreSinkOperator` was used for operations on cached tables. Now, it's only used for operations on cached tables that are not unified (i.e., just stored in memory).
  For operations on unified views, query plans will use a `FileSinkOperator` that's modified by the `OperatorFactory` to use a disk SerDe instead of `ColumnarSerDe`, since the operator's input rows may be lazy objects.
- New SparkLoadTask
  <p> For `CREATE TABLE`, `CTAS`, or `LOAD ... OVERWRITE INTO`, the execution process starts with a `HadoopTableReader` that scans tables or partitions data directories from the Hive warehouse into a Hadoop RDD. That RDD is then transformed and materialized it into an `RDD[TablePartition]` backed by columnar store.
  For `INSERT INTO`, or `LOAD ... INTO` append updates, the `HadoopTableReader` uses a path filter that will only accept new files loaded into the target table/partition data directory by Hive's `FileSinkOperator` and `MoveTask`. That path filter is created using a snapshot of the target directory during semantic analysis.


harveyfeng:
Updates:
- `TableRecovery` added for reloading any RDD with the "shark.cache.reloadOnRestart" flag. That flag defaults to 'true' for any unified view/cached table creation. `TableRecovery` also replaces the old `CachedTableRecovery` that executed the CTAS string on startup.
- `TblProperties` counterpart to `SharkConfVars`
- Simple rewrite util for `CACHE src` and `UNCACHE src` commands

rxin:
you guys should take a look at this too

@jhartlaub @markhamstra

src/main/scala/shark/SharkEnv.scala
```diff
@@ -17,17 +17,23 @@
 
 package shark
 
+import scala.collection.JavaConversions._
```
rxin: Most of the additional imports in this file are not used. Can you remove them?

src/main/scala/shark/execution/HadoopTableReader.scala
```diff
@@ -53,14 +55,22 @@ class HadoopTableReader(@transient _tableDesc: TableDesc, @transient _localHConf
...
+      _tableDesc.getDeserializerClass.asInstanceOf[Class[Deserializer]])
+
   /**
    * Creates a Hadoop RDD to read data from the target table's data directory. Returns a transformed
    * RDD that contains deserialized rows.
    */
-  def makeRDDForTable(hiveTable: HiveTable, filterOpt: Option[PathFilter] = None): RDD[_] = {
+  def makeRDDForTable(
+      hiveTable: HiveTable,
+      filterOpt: Option[PathFilter],
```
rxin, line 76: Can you add to the comment what filterOpt does and how it is used? Also do it for makeRDDForPartitionedTable.

src/main/scala/shark/execution/RDDUtils.scala
```diff
@@ -33,16 +33,22 @@ import shark.SharkEnv
...
   def unionAndFlatten[T: ClassManifest](
     rdd: RDD[T],
-    otherRdd: RDD[T]): RDD[T] = {
-    val unionedRdd = otherRdd match {
-      case unionRdd: UnionRDD[_] => {
-        new UnionRDD(rdd.context, unionRdd.rdds :+ rdd)
-      }
-      case _ => rdd.union(otherRdd)
+    otherRdd: RDD[T]): UnionRDD[T] = {
+    val otherRdds = otherRdd match {
```
rxin: Explicitly declare the type of otherRdds and rdds here.

rxin:
will continue the reviewing tomorrow. a note to myself: "unified view" isn't exactly a term we want to expose to the users ...

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+import org.apache.spark.storage.StorageLevel
+
+import shark.{LogHelper, SharkEnv, Utils}
+import shark.execution.serialization.KryoSerializer
+import shark.memstore2._
+import shark.util.HiveUtils
+
+
+private[shark]
+class SparkLoadWork(
```
rxin, line 60: Add some javadoc explaining the parameters, in particular, commandType, unifyView, and reloadOnRestart.

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+    val commandType: SparkLoadWork.CommandTypes.Type,
+    val preferredStorageLevel: StorageLevel,
+    val cacheMode: CacheType.CacheType,
+    val unifyView: Boolean,
+    val reloadOnRestart: Boolean)
+  extends java.io.Serializable {
+
+  var pathFilterOpt: Option[PathFilter] = None
+
+  var partSpecs: Seq[JavaMap[String, String]] = Nil
```
rxin, line 87: Comment on what the mapping is (what does the first string mean, and what does the 2nd string mean)

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+    // Not the most efficient, but this method isn't called very often - either a single partition
+    // spec is passed for partition update, or all partitions are passed for cache load operation.
+    partSpecs = partSpecs ++ Seq(partSpec)
+  }
+}
+
+object SparkLoadWork {
+  object CommandTypes extends Enumeration {
+    type Type = Value
+    val OVERWRITE, INSERT, NEW_ENTRY = Value
```
rxin, line 108: Can you also comment on what these 3 values mean, and example queries that would trigger them?

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+private[shark]
+class SparkLoadWork(
+    val databaseName: String,
+    val tableName: String,
+    val commandType: SparkLoadWork.CommandTypes.Type,
+    val preferredStorageLevel: StorageLevel,
+    val cacheMode: CacheType.CacheType,
+    val unifyView: Boolean,
+    val reloadOnRestart: Boolean)
+  extends java.io.Serializable {
```
rxin: Remove java.io

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+ * LOAD INTO/OVERWRITE, CACHE, and CTAS commands.
+ */
+private[shark]
+class SparkLoadTask extends HiveTask[SparkLoadWork] with Serializable with LogHelper {
+
+  override def execute(driveContext: DriverContext): Int = {
+    logDebug("Executing " + this.getClass.getName)
+
+    // Set Spark's job description to be this query.
+    SharkEnv.sc.setJobDescription("Loading from Hadoop for a(n) " + work.commandType)
```
rxin: What is a(n)? We need to change the log message since data is not always from Hadoop. Maybe just say "Loading table X into memory" (or table X parititon ...)

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+   * Handles loading data from disk into the Shark cache for non-partitioned tables.
+   *
+   * @hiveTable Hive metadata object representing the target table.
+   * @hadoopReader Used to create a HadoopRDD from the table's data directory.
+   * @pathFilterOpt Defined for INSERT update operations (e.g., INSERT INTO) and passed to
+   *     hadoopReader#makeRDDForTable() to determine which new files should be read from the table's
+   *     data directory - see the SparkLoadWork#apply() factory method for an example of how a
+   *     path filter is created.
+   */
+  def loadMemoryTable(
```
rxin: Let's make the naming consistent between loadMemoryTable and loadPartitionedTable. Either use memory in both or not.

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+      serDe.getClass)
+    // Transform the HadoopRDD to an RDD[TablePartition].
+    val (tablePartitionRDD, tableStats) = transformAndMaterializeInput(
+      inputRDD,
+      tableSchema,
+      memoryTable.preferredStorageLevel,
+      hadoopReader.broadcastedHiveConf,
+      serDe.getObjectInspector.asInstanceOf[StructObjectInspector])
+    memoryTable.tableRDD = work.commandType match {
+      case (SparkLoadWork.CommandTypes.OVERWRITE
```
rxin:
```
case case (SparkLoadWork.CommandTypes.OVERWRITE | SparkLoadWork.CommandTypes.NEW_ENTRY) =>
  tablePartitionRDD
```

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+          case None => Unit
+        }
+        unionedRDD
+      }
+    }
+    SharkEnv.memoryMetadataManager.putStats(databaseName, tableName, tableStats.toMap)
+  }
+
+  /**
+   * Returns Shark PartitionedMemorytable that was created or fetched from the metastore, based on
```
rxin: Explain in what scenarios will this "create" a new table, or just "get"

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+   * the command type handled by this task.
+   *
+   * @hiveTable Corresponding HiveTable for the Shark PartitionedMemorytable that will be returned.
```
rxin: Add pathSpecs also to the javadoc and again explain what the 1st string means, and what the 2nd string means.

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+          classOf[ColumnarSerDe].getName,
+          conf)
+        newPartitionedTable
+      }
+      case _ => {
+        SharkEnv.memoryMetadataManager.getTable(databaseName, tableName) match {
+          case Some(table: PartitionedMemoryTable) => table
+          case _ => {
+            throw new Exception(
+              "Internal error: cached, partitioned table for INSERT handling doesn't exist.")
```
rxin: Maybe use an assert?

i.e.

```
val table =  SharkEnv.memoryMetadataManager.getTable(databaseName, tableName)
assert(table.isDefined && table.get isInstanceOf ...)
table
```

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+        hadoopReader.broadcastedHiveConf,
+        unionOI)
+      // Determine how to cache the table RDD created.
+      val tableOpt = partitionedTable.getPartition(partitionKey)
+      if (tableOpt.isDefined && (work.commandType == SparkLoadWork.CommandTypes.INSERT)) {
+        val previousRDD = tableOpt.get
+        partitionedTable.updatePartition(
+          partitionKey, RDDUtils.unionAndFlatten(tablePartitionRDD, previousRDD))
+        // Union stats for the previous RDD with the new RDD loaded.
+        SharkEnv.memoryMetadataManager.getStats(databaseName, tableName) match {
```
rxin: Do we ever expect the None code path being triggered? If not, let's just use 

```
SparkLoadTask.unionStatsMaps(
  tableStats,
  SharkEnv.memoryMetadataManager.getStats(databaseName, tableName).get)
```

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+    baseSerDeProps: Properties): Properties = {
+    val serDeProps = new Properties(baseSerDeProps)
+    // Delimited by ","
+    var columnNameProperties: String = serDeProps.getProperty(Constants.LIST_COLUMNS)
+    // `None` if column types are missing. By default, Hive SerDeParameters initialized by the
+    // ColumnarSerDe will treat all columns as having string types.
+    // Delimited by ":"
+    var columnTypePropertiesOpt = Option(serDeProps.getProperty(Constants.LIST_COLUMN_TYPES))
+
+    for (partColName <- partCols) {
```
rxin: Can we rewrite this using mkString?

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+      columnNameProperties += "," + partColName
+    }
+    serDeProps.setProperty(Constants.LIST_COLUMNS, columnNameProperties)
+    columnTypePropertiesOpt.foreach { columnTypeProperties =>
+      var newColumnTypeProperties = columnTypeProperties
+      for (partColName <- partCols) {
```
rxin: Isn't this simply

```
(":" + Constants.STRING_TYPE_NAME) * partCols.size
```

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+   * Returns a copy of `baseSerDeProps` with the names and types for the table's partitioning
+   * columns appended to respective row metadata properties.
+   */
+  def addPartitionInfoToSerDeProps(
+    partCols: Seq[String],
+    baseSerDeProps: Properties): Properties = {
+    val serDeProps = new Properties(baseSerDeProps)
+    // Delimited by ","
+    var columnNameProperties: String = serDeProps.getProperty(Constants.LIST_COLUMNS)
+    // `None` if column types are missing. By default, Hive SerDeParameters initialized by the
```
rxin, line 426: Would be great to add some examples in the comment on what the serde props expect (instead of simply saying delimited by "," and ":"

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,457 @@
...
+      var newColumnTypeProperties = columnTypeProperties
+      for (partColName <- partCols) {
+        newColumnTypeProperties += ":" + Constants.STRING_TYPE_NAME
+      }
+      serDeProps.setProperty(Constants.LIST_COLUMN_TYPES, newColumnTypeProperties)
+    }
+    serDeProps
+  }
+
+  def unionStatsMaps(
```
rxin: Maybe rename this to appendStatsMap and remove the return type? The current name implies it is some immutable operation.

rxin: That got me thinking we should really just store an array for the stats, and use the array's index as the partition index. It's fine for this pull request to leave it as is, but if it makes your code simpler, you should consider doing it as part of this PR too.

src/main/scala/shark/parse/QueryBlock.scala
```diff
@@ -31,11 +35,41 @@ class QueryBlock(outerID: String, alias: String, isSubQuery: Boolean)
...
+
+  private var _reloadOnRestart: Boolean = false
+
+  private var _preferredStorageLevel: StorageLevel = StorageLevel.NONE
+
+  // Whether the created to be created or the table specified by CACHED should be backed by disk.
+  private var _unifyView = false
+
+  // Descriptor for the table being updated by an INSERT.
+  private var _targetTableDesc: TableDesc = _
```
rxin: If we have both a setter and a getter, let's just remove the private part and name this targetTableDesc. Similarly, unifyView, cacheModeForCreateTable, etc.

src/main/scala/shark/execution/OperatorFactory.scala
```diff
@@ -22,6 +22,7 @@ import scala.collection.JavaConversions._
 import org.apache.hadoop.hive.ql.exec.{GroupByPostShuffleOperator, GroupByPreShuffleOperator}
 import org.apache.hadoop.hive.ql.exec.{Operator => HOperator}
 import org.apache.hadoop.hive.ql.metadata.HiveException
+import org.apache.hadoop.hive.serde2.Deserializer
```
rxin: Remove this and then this file disappears from the change list.

src/main/scala/shark/execution/SharkDDLTask.scala
```diff
@@ -77,15 +78,17 @@ private[shark] class SharkDDLTask extends HiveTask[SharkDDLWork]
   /** Handles a CREATE TABLE or CTAS. */
   def createTable(
```
rxin, line 79: Is this metadata operation only? If yes, add that to the javadoc.

src/main/scala/shark/execution/SharkDDLTask.scala
```diff
@@ -77,15 +78,17 @@ private[shark] class SharkDDLTask extends HiveTask[SharkDDLWork]
...
     val tableName = createTblDesc.getTableName
     val tblProps = createTblDesc.getTblProps
-
     val preferredStorageLevel = MemoryMetadataManager.getStorageLevelFromString(
       tblProps.get("shark.cache.storageLevel"))
+    val unifyView = tblProps.get(SharkTblProperties.UNIFY_VIEW_FLAG.varname).toBoolean
+    val reloadOnRestart = tblProps.get(SharkTblProperties.RELOAD_ON_RESTART_FLAG.varname).toBoolean
     val isHivePartitioned = (createTblDesc.getPartCols.size > 0)
-    if (isHivePartitioned) {
+    val newTable = if (isHivePartitioned) {
```
rxin: This newTable is not used. Is that intentional? If yes, let's remove the variable.

src/main/scala/shark/execution/HadoopTableReader.scala
```diff
@@ -93,18 +107,27 @@ class HadoopTableReader(@transient _tableDesc: TableDesc, @transient _localHConf
     deserializedHadoopRDD
   }
 
-  override def makeRDDForPartitionedTable(partitions: Seq[HivePartition]): RDD[_] =
-    makeRDDForPartitionedTable(partitions, filterOpt = None)
+  override def makeRDDForPartitionedTable(
+      partitions: Seq[HivePartition]): RDD[_] = {
+    val partitionToDeserializer = partitions.map(part =>
+      (part, part.getDeserializer.getClass.asInstanceOf[Class[Deserializer]])).toMap
+    makeRDDForPartitionedTable(partitionToDeserializer, None /* filterOpt */)
```
markhamstra, line 114: No real value in explicitly calling with the default value, so instead just `makeRDDForPartitionedTable(partitionToDeserializer)`

markhamstra, line 114: And not only do I prefer the prior existence of the default, but I'd also rather see the prior convention of `filterOpt = Some(...)` instead of `Some(...) /* filterOpt */`

src/main/scala/shark/execution/RDDUtils.scala
```diff
@@ -33,16 +33,22 @@ import shark.SharkEnv
...
-      }
-      case _ => rdd.union(otherRdd)
+    otherRdd: RDD[T]): UnionRDD[T] = {
+    val otherRdds: Seq[RDD[T]] = otherRdd match {
+      case otherUnionRdd: UnionRDD[_] => otherUnionRdd.rdds
+      case _ => Seq(otherRdd)
+    }
+    val rdds: Seq[RDD[T]] = rdd match {
+      case unionRdd: UnionRDD[_] => unionRdd.rdds
+      case _ => Seq(rdd)
```
markhamstra, line 49: Could be DRYer

harveyfeng:
Closing this one for now. I'll submit a new PR with additional changes to simplify caching semantics:
- Rely on a `shark.cache` table property, which can have to following values:
  - `memory`: table will be stored in memory and on disk (cache is write-through). Persistent across Shark sessions. By default, a Shark session will reload all previously cached tables into memory.
  - `memory-only`: table will only be stored in memory, and dropped at the end of a session.
  - `tachyon`
  - `none`: normal Hive table

This replaces the `shark.cache.storageLevel` and `shark.cache.unifyView` properties available to the user ATM.

src/main/scala/shark/execution/SparkLoadTask.scala
```diff
@@ -0,0 +1,449 @@
...
+  /**
+   * Factory/helper method used in LOAD and INSERT INTO/OVERWRITE analysis. Sets all necessary
+   * fields in the SparkLoadWork returned.
+   */
+  def apply(
+      db: Hive,
+      conf: HiveConf,
+      hiveTable: HiveTable,
+      partSpecOpt: Option[JavaMap[String, String]],
+      isOverwrite: Boolean): SparkLoadWork = {
```
markhamstra, line 120: All of the repetition of `SparkLoadWork.CommandTypes.` gets tedious.  I'd `import CommandTypes._` here and `import SparkLoadWork.CommandTypes._` right after `class SparkLoadTask ... {`, then just use `OVERWRITE` and `INSERT` throughout this file -- e.g., I'd go with `val commandType = if (isOverwrite) OVERWRITE else INSERT`.

markhamstra:
@rxin Have you established error-handling practices/conventions?  I question whether you really want to use the several instances of this pattern from this pull request:

``` scala
val somethingOpt = someSource.getSomething(...)
assert (somethingOpt.exists(expectation), "Didn't get the expected something.")
somethingOpt.get
```

src/main/scala/shark/memstore2/TableRecovery.scala
```diff
@@ -0,0 +1,59 @@
...
+   * @param cmdRunner The runner that is responsible for taking a cached table query and
+   *        a) Creating the table metadata in Hive Meta Store
+   *        b) Loading the table as an RDD in memory
+   *        @see SharkServer for an example usage.
+   */
+  def loadUnifiedViews(cmdRunner: String => Unit) {
+    // Filter for tables that should be reloaded into the cache.
+    val currentDbName = db.getCurrentDatabase()
+    for (databaseName <- db.getAllDatabases()) {
+      for (tableName <- db.getAllTables(databaseName)) {
```
markhamstra, line 45: Why nested `for` instead of:

``` scala
for {
  databaseName <- db.getAllDatabases
  tableName <- db.getAllTables(databaseName)
  ...
}
```

src/main/scala/shark/parse/QueryBlock.scala
```diff
@@ -28,14 +30,22 @@ import shark.memstore2.CacheType._
  * and analyzing ASTs (e.g. in SharkSemanticAnalyzer#analyzeCreateTable()).
  */
 class QueryBlock(outerID: String, alias: String, isSubQuery: Boolean)
-    extends HiveQueryBlock(outerID, alias, isSubQuery) {
+  extends HiveQueryBlock(outerID, alias, isSubQuery) {
 
   // The CacheType for the table that will be created from CREATE TABLE/CTAS.
-  private var cacheModeForCreateTable = CacheType.NONE
+  var cacheModeForCreateTable = CacheType.NONE
```
markhamstra, line 36: `private[parse] var cacheModeForCreateTable = CacheType.NONE`?

markhamstra, line 36: And do the other mutable fields of `QueryBlock` need to be public?

harveyfeng, line 36: They do - it's used for planning in SharkSemanticAnalyzer

harveyfeng, line 36: Oh, sorry, missed the `private [parse]` from above.
Not sure how stringent the access modifiers right now should be - most other Shark code hasn't focused on it. Given that Shark and related components are undergoing pretty active development right now, it might be worth holding off until (much...) later, when the codebase stabilizes and we can make a batch access-modifier PR for each package in the codebase.

src/main/scala/shark/execution/SparkLoadTask.scala
harveyfeng, line 120: I was hoping to keep the full specification for each of those types, since they're properties of the SparkLoadWork descriptor that are set during planning. This way, each mention of `SparkLoadWork.CommandTypes.INSERT/OVERWRITE` in the SparkLoadTask code makes it clear that the execution phase is responding to what was planned and put into SparkLoadWork.

harveyfeng:
@markhamstra @rxin 
There is a QueryExecptionException in the `shark.api` package right now that's reserved for exceptions caused by the user. Assertions throughout the codebase to enforce internal invariants seem to be fit with conventional Scala style.

src/main/scala/shark/parse/QueryBlock.scala
markhamstra, line 36: The too-big public interfaces of Spark and Shark are something that has to change if they are to graduate to the status of fully production-ready code.  Making life easier for Spark and Shark developers by leaving everything wide-open makes life a lot harder for Spark and Shark users.

harveyfeng:
Btw, thanks for the review @markhamstra! Going to submit an updated, cleaner PR once unit tests are cleared for it.

PR #217, opened 2013-11-18
CancelJob through JDBC Clients 

sundeepn:
Handle disconnects relies on TCP keepalives (To be set using OS settings on server)
Allows explicit kill session/job groups through JDBC.
(Local mode job cleanup already handled)

### DIFF
--- a/src/main/scala/shark/SharkCliDriver.scala
+++ b/src/main/scala/shark/SharkCliDriver.scala
@@ -43,7 +43,7 @@ import org.apache.hadoop.hive.ql.processors.{CommandProcessor, CommandProcessorF
 import org.apache.hadoop.hive.ql.session.SessionState
 import org.apache.hadoop.hive.shims.ShimLoader
 import org.apache.hadoop.io.IOUtils
-import org.apache.thrift.transport.TSocket;
+import org.apache.thrift.transport.TSocket
 
 
 
@@ -69,7 +69,7 @@ object SharkCliDriver {
         } else {
           if (transport != null) {
             // Force closing of TCP connection upon session termination
-        	transport.getSocket().close()
+            transport.getSocket().close()
           }
         }
       }
--- a/src/main/scala/shark/SharkCliDriver.scala
+++ b/src/main/scala/shark/SharkCliDriver.scala
@@ -43,13 +43,16 @@ import org.apache.hadoop.hive.ql.processors.{CommandProcessor, CommandProcessorF
 import org.apache.hadoop.hive.ql.session.SessionState
 import org.apache.hadoop.hive.shims.ShimLoader
 import org.apache.hadoop.io.IOUtils
+import org.apache.thrift.transport.TSocket;
+
 
 
 object SharkCliDriver {
 
   private var prompt  = "shark"
   private var prompt2 = "     " // when ';' is not yet seen.
-
+  private var transport:TSocket = _
+    
   installSignalHandler()
 
   /**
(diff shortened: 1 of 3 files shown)


rxin:
Jenkins, test this please.

src/main/scala/shark/SharkServer.scala
```diff
@@ -145,6 +165,26 @@ object SharkServer extends LogHelper {
       }
     }
   }
+  
+  class JDBCWatcher(sock:java.net.Socket, sessionID:String) extends Thread {
+    
+	override def run() {
```
rxin: Can you indent this properly?

src/main/scala/shark/SharkCliDriver.scala
```diff
@@ -60,7 +63,15 @@ object SharkCliDriver {
...
     HiveInterruptUtils.add(new HiveInterruptCallback {
       override def interrupt() {
-        SharkEnv.sc.cancelAllJobs()
+        // Handle remote execution mode
+        if (SharkEnv.sc != null) {
+          SharkEnv.sc.cancelAllJobs()
+        } else {
+          if (transport != null) {
+            // Force closing of TCP connection upon session termination
+        	transport.getSocket().close()
```
rxin: Fix the indent here.

src/main/scala/shark/SharkCliDriver.scala
```diff
@@ -43,13 +43,16 @@ import org.apache.hadoop.hive.ql.processors.{CommandProcessor, CommandProcessorF
 import org.apache.hadoop.hive.ql.session.SessionState
 import org.apache.hadoop.hive.shims.ShimLoader
 import org.apache.hadoop.io.IOUtils
+import org.apache.thrift.transport.TSocket;
```
rxin: Remove the trailing ;

src/main/scala/shark/SharkServer.scala
```diff
@@ -85,8 +87,26 @@ object SharkServer extends LogHelper {
     serverTransport = new TServerSocket(cli.port)
 
     val hfactory = new ThriftHiveProcessorFactory(null, new HiveConf()) {
-      override def getProcessor(t: TTransport) =
-        new ThriftHive.Processor(new GatedSharkServerHandler(latch))
+      override def getProcessor(t: TTransport) = {
+        var remoteClient = "Unknown"
```
rxin, line 91: Let's use a combination of address + port, instead of address + time.

sundeepn, line 91: Looks like we are always connected on the JDBC port on the server side. Is there a reason you want to use the port?

rxin, line 91: I thought the port was the client port, not the server port. I am just thinking it is a little bit fragile to use address + time because the same machine might (although very small chance) create two clients at the same time. But if there is no way to get the client port, that's fine.

sundeepn, line 91: We can add a random number to be really sure.

sundeepn, line 91: I will add the remote port for now.

src/main/scala/shark/SharkServer.scala
```diff
@@ -85,8 +87,28 @@ object SharkServer extends LogHelper {
...
+        if (t.isInstanceOf[TSocket]) {
+          remoteClient = t.asInstanceOf[TSocket].getSocket()
+            .getRemoteSocketAddress().asInstanceOf[InetSocketAddress]
+            .getAddress().toString()
+        }
+        logInfo("Audit Log: Connection Initiated with JDBC client - " + remoteClient)
+        
+        val jdbcSocket = t.asInstanceOf[TSocket].getSocket()
```
rxin: Should this part also go into the if body? Otherwise there is no point checking whether t is an instance of TSocket earlier. Other than that, this looks good!

rxin:
Thanks. Merged this.

PR #220, opened 2013-12-01
AST transformations for DISTINCT aggregates

harveyfeng:
#### Recap

For queries with DISTINCT aggregates that don’t have a GROUP BY (i.e., the query returns a single number), Shark relies on a single shuffle and reduce for both processing the distincts (using an in-memory sort) and the final aggregation. THis is a pretty severe bottleneck

Queries with GROUP BY aren’t an issue because full aggregation for each key can be done locally on a reducer after a hash-partitioned shuffle.

Note: This type of optimization will be moved to the optimizer level in a future dev cycle. For a preview, check out the Catalyst abstract ([link](http://spark-summit.org/talk/armbrust-catalyst-a-query-optimization-framework-for-spark-and-shark/)) and the corresponding slides ([download link](http://spark-summit.org/wp-content/uploads/2013/10/J-Michael-Armburst-catalyst-spark-summit-dec-2013.pptx))
#### Approach

Execution for DISTINCT aggregates (without GROUP BY) should be split into two shuffles: the first to hash-partition selected expressions and the second to rendezvous at a single reducer that will do the final aggregation.

<i> In the diagrams below, the nodes in red are modified and the nodes in blue are moved from the original AST. Nodes in green are directly instantiated in Shark. Nodes not colored remain unchanged. </i>

In the driver, `ASTRewriteUtil#countDistinctToGroupBy()` searches for patterns of the AST representation for

```
SELECT <aggregate fn>(DISTINCT <exprs>) FROM <source> <WHERE, ORDER BY, etc>
```

where `<source>` is a table, subquery, or a table or subquery alias.
countdistinct 1

Then, for each subtree that matches the pattern, transform it to the AST representation for the rewriten query, 

```
SELECT <aggregate fn>(*) FROM
   (SELECT DISTINCT <exprs> FROM <source> <WHERE, ORDER BY, etc>) <Shark-internal alias>
```

countdistinctrewrite 1

### DIFF
--- a/src/main/scala/shark/SharkDriver.scala
+++ b/src/main/scala/shark/SharkDriver.scala
@@ -37,7 +37,7 @@ import shark.api.TableRDD
 import shark.api.QueryExecutionException
 import shark.execution.{SharkExplainTask, SharkExplainWork, SparkTask, SparkWork}
 import shark.memstore2.ColumnarSerDe
-import shark.parse.{QueryContext, SharkSemanticAnalyzerFactory}
+import shark.parse.{QueryContext, SharkExplainSemanticAnalyzer, SharkSemanticAnalyzerFactory}
 
 
 /**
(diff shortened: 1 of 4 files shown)


aarondav:
For posterity, this PR is intended to solve what issue with the prior DISTINCT aggregations? Is that problem a non-issue if there is a GROUP BY?

harveyfeng:
The issue's in queries with DISTINCT aggregates that don’t have a GROUP BY (i.e., the query returns a single number). In this case, Shark relies on a single reducer for both processing the distincts (using an in-memory sort) and the final aggregation, which is a pretty severe bottleneck

If there's a GROUP BY, full count distincts for each key are (already) done on the slave that the key hash-partitioned to.

harveyfeng:
This lone udf_count failure only has to do with EXPLAIN...

rxin:
Thanks. Merging this into branch-0.8.

PR #221, opened 2013-12-09
CG (Row/OI) & Example for SelectOperator

chenghao-intel:
Hi, All
    This is the first complete version for CodeGen(CG) Row / ObjectInspector (OI), with all Hive data types support; and also I put the the codegen of SelectOperator to demonstrate how to integrate the CGRow / CGOI with the Operators.
1. Where to Start (add the jvm flag -Dcg_source_code_path=/home/hcheng/git/sotc_cloud-shark/cgjava)
   a. Run unit test shark.execution.cg.TestCGRowOI
   b. Run Query like "select map(id, struct(name, id, if(ds='2011-10-23', 123, null))) from things;" (set shark.query.cg=false; to disable the codegen feature)
2. TODOs
   a. The source code generation utilizes the ScalaTE (Scala Template Engine), all of the templates are now compiling in runtime, but could be precompiled in packing the binary, need to figure out how to do that in SBT.

PS:
ScalaTE tags reference: http://scalate.fusesource.org/documentation/ssp-reference.html


rxin:
Thanks! Will take a look at this soon.

rxin:
Jenkins, add to whitelist.

PR #226, opened 2014-01-03
Add -nc to wget to avoid it "not clobber" (or download a *.n version of) the file if it already exists.

andyk:
I noticed the test in jenkins has re-downloaded this 49 times now... https://amplab.cs.berkeley.edu/jenkins/job/Shark-Master-Hadoop1/ws/run-tests-from-scratch-workspace/

### DIFF
--- a/bin/dev/run-tests-from-scratch
+++ b/bin/dev/run-tests-from-scratch
@@ -226,7 +226,7 @@ fi
 # Download Scala if SCALA_HOME is not specified.
 ####################################################################
 if [ "x$SCALA_HOME" == "x" ] ; then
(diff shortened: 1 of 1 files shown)


andyk:
Even better than (and instead of) using "-nc", we should just remove past downloads of this, because otherwise if the script is run a 2nd time with an updated value of $SCALA_DOWNLOAD_PATH then the expression for untar-ing might expand to be something like `tar xvfz scala-2.9.3.tgz scala-2.10.0.tgz` which fails.

rxin:
@andyk  It this pr still active?

andyk:
Yeah, i think this is worth committing.

PR #233, opened 2014-01-06
SHARK-210: Add ROLLUP/CUBE/GROUPING SET feature to Shark

sun-rui:
ROLLUP/CUBE/GROUPING SET are extensions to GROUPBY which are useful in generating summary tables. They have been supported since HIVE 0.10.

This patch is to add support for ROLLUP/CUBE/GROUPING SET to Shark on Shark's hive0.11 branch.


rxin:
Jenkins, add to whitelist.

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
```diff
@@ -104,6 +114,50 @@ class GroupByPreShuffleOperator extends UnaryOperator[HiveGroupByOperator] {
...
+        groupingSetsBitSet += new BitSet1(groupingSet.longValue())
+      }
+    }
+  }
+
+  // Return an iterator which iterates newKeys corresponding to the grouping sets.
+  protected final
+  def getNewKeysIterator (newKeysArray: Array[Object]): Iterator[Unit] = {
+    new Iterator[Unit]() {
+      for (i <- 0 until groupingSetsPosition) {
```
rxin: You can replace this with an array copy - much faster also.

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
```diff
@@ -104,6 +114,50 @@ class GroupByPreShuffleOperator extends UnaryOperator[HiveGroupByOperator] {
...
+        cloneNewKeysArray(i) = newKeysArray(i)     
+      }
+      private var groupingSetIndex = 0
```
rxin: this is private by default anyway so no need to specify private here.

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
```diff
@@ -104,6 +114,50 @@ class GroupByPreShuffleOperator extends UnaryOperator[HiveGroupByOperator] {
...
+
+      override def hasNext: Boolean =
+        if (groupingSetIndex < groupingSets.size()) true else false
```
rxin: This can be simplified to

```
groupingSetIndex < groupingSets.size
```

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
```diff
@@ -104,6 +114,50 @@ class GroupByPreShuffleOperator extends UnaryOperator[HiveGroupByOperator] {
...
+
+        newKeysGroupingSets += groupingSetValueEvaluator.evaluate(null)
+        groupingSetsBitSet += new BitSet1(groupingSet.longValue())
+      }
+    }
+  }
+
+  // Return an iterator which iterates newKeys corresponding to the grouping sets.
+  protected final
+  def getNewKeysIterator (newKeysArray: Array[Object]): Iterator[Unit] = {
```
rxin, line 138: Can you add some comments on how this function is used and why it exists?

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
```diff
@@ -104,6 +114,50 @@ class GroupByPreShuffleOperator extends UnaryOperator[HiveGroupByOperator] {
     keyObjectInspector = ObjectInspectorFactory.getStandardStructObjectInspector(keyFieldNames, keyois)
 
     keyFactory = new KeyWrapperFactory(keyFields, keyObjectInspectors, currentKeyObjectInspectors)
+    
+    // Initializations for grouping set.
+    groupingSetsPresent = conf.isGroupingSetsPresent()
+    if (groupingSetsPresent) {
+      groupingSets = conf.getListGroupingSets()
+      groupingSetsPosition = conf.getGroupingSetPosition()
+      newKeysGroupingSets = new ArrayBuffer[Object]()
```
rxin:
```
newKeysGroupingSets = new Array[Object](groupingSets.size)
groupingSetsBitSet = new Array[BitSet1](groupingSets.size)
```

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
```diff
@@ -61,6 +63,14 @@ class GroupByPreShuffleOperator extends UnaryOperator[HiveGroupByOperator] {
 
   @transient var aggregationIsDistinct: Array[Boolean] = _
 
+  // Grouping set related properties.
+  @transient var groupingSetsPresent: Boolean = _
+  @transient var groupingSets: java.util.List[java.lang.Integer] = _
+  @transient var groupingSetsPosition: Int = _
+  @transient var newKeysGroupingSets: ArrayBuffer[Object] = _
```
rxin: Maybe I'm mistaken, but I think you can use Array directly for these two, and it will be faster also. 

BTW - I'm curious, how come you are using BitSet1?

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
```diff
@@ -122,22 +176,28 @@ class GroupByPreShuffleOperator extends UnaryOperator[HiveGroupByOperator] {
...
-      if (isNewKey) {
-        aggregateNewKey(row, aggs)
-      } else {
-        aggregateExistingKey(row, aggs)
-      }
+      val newKeysIter =
+        if (groupingSetsPresent) getNewKeysIterator(newKeys.getKeyArray) else null
+
+      do {
+        if (groupingSetsPresent) newKeysIter.next
```
rxin: Use curly braces even for one-liners, and add parenthesis for next since it has side-effects. i.e.

``` scala
if (groupingSetsPresent) {
  newKeysIter.next()
}
```

src/main/scala/shark/execution/GroupByPostShuffleOperator.scala
```diff
@@ -437,7 +437,25 @@ class GroupByPostShuffleOperator extends GroupByPreShuffleOperator with HiveTopO
     if (!newIter.hasNext && keyFields.length == 0) {
       Iterator(createEmptyRow()) // We return null if there are no rows
     } else {
-      newIter
+      if (groupingSetsPresent) {
+        // An additional MR job is introduced since the cardinality of grouping sets
```
rxin: Perhaps update the comment since we are no longer running "MR".

rxin:
Thanks Rui! I left some comments. This will be really cool to have.

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
sun-rui: groupingSets is a set of bitmask of which a set bit represents a groupby key which is an element of the corresponding group set. BitSet1 is used for quick access to the groupby keys in a grouping set, similar to the HIVE implementation. If you think bit testing via shift operation is better, I can remove it.

rxin: When is the bit set in the BitSet? I didn't find it in the diff (maybe I missed it)

src/main/scala/shark/execution/GroupByPreShuffleOperator.scala
sun-rui, line 138: This iterator abstracts the operation that gets an array of groupby keys for the next grouping set from the full set of grouping sets and makes such logic re-usable in several places. Actually, a more generic iterator could be defined to hide the differences between presence and absence of grouping sets, which would make the code change simpler and cleaner. But for performance consideration, separate handling of presence and absence of grouping sets was chosen.

sun-rui:
I ran HIVE CLI Tests with the following command:
TEST=groupby_grouping_set.\* ./bin/dev/test

Actually all groupby_grouping_set*.q have correct results, but groupby_grouping_sets3.q/groupby_grouping_sets4.q/groupby_grouping_sets5.q were reported failed due to an issue in HIVE 0.11 for Shark. The issue is that explain information for queries in the 3 q.out files are out-of-sync.

Should I fix this issue and add groupby_grouping_set*.q into src/test/tests_pass.txt?

rxin:
This PR looks pretty good now. Thanks for updating this! Let's add the passing tests to the whitelist first, and then it would be great if you can submit a separate pull request to address the failing tests.

rxin:
Thanks. I've merged this!

PR #234, opened 2014-01-09
Fix SerializationSuite failures due to SharkContext NPE

harveyfeng:
Changes:
- SharkEnv always contains a reference to SharkContext type (as opposed to SparkContext)
- Added initialize() methods to Shark's KryoSerializer singleton, to make it easier to pass SparkConf references needed for SparkKryoSerializer
- Always launch with Java for `bin/shark-shell`, due to scala-2.10's `scala` command loading a conflicting version of Akka.


harveyfeng:
Missed an import in the previous version

rxin:
Thanks!

PR #245, opened 2014-01-17
fix KryoSerializer NPE.

ChengXiangLi:
pull request #234 introduce an KryoSerializer NPE due to ser is not initialized in spark task,

```
java.lang.NullPointerException
        at shark.execution.serialization.KryoSerializer$.deserialize(KryoSerializer.scala:49)
        at shark.execution.serialization.OperatorSerializationWrapper.value(OperatorSerializationWrapper.scala:50)
        at shark.execution.package$.opSerWrapper2op(package.scala:30)
        at shark.execution.Operator$$anonfun$executeProcessPartition$1.apply(Operator.scala:336)
        at shark.execution.Operator$$anonfun$executeProcessPartition$1.apply(Operator.scala:335)
        at org.apache.spark.rdd.RDD$$anonfun$2.apply(RDD.scala:460)
        at org.apache.spark.rdd.RDD$$anonfun$2.apply(RDD.scala:460)
        at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:34)
        at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:241)
        at org.apache.spark.rdd.RDD.iterator(RDD.scala:232)
        at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:109)
        at org.apache.spark.scheduler.Task.run(Task.scala:53)
        at org.apache.spark.executor.Executor$TaskRunner$$anonfun$run$1.apply$mcV$sp(Executor.scala:213)
        at org.apache.spark.deploy.SparkHadoopUtil.runAsUser(SparkHadoopUtil.scala:50)
        at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:178)
        at java.util.concurrent.ThreadPoolExecutor$Worker.runTask(ThreadPoolExecutor.java:886)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:908)
        at java.lang.Thread.run(Thread.java:662)
```

### DIFF
--- a/src/main/scala/shark/execution/serialization/KryoSerializer.scala
+++ b/src/main/scala/shark/execution/serialization/KryoSerializer.scala
@@ -19,7 +19,7 @@ package shark.execution.serialization
 
 import java.nio.ByteBuffer
 
-import org.apache.spark.SparkConf
+import org.apache.spark.{SparkEnv, SparkConf}
 import org.apache.spark.serializer.{KryoSerializer => SparkKryoSerializer}
 
 import shark.SharkContext
@@ -31,7 +31,7 @@ import shark.SharkContext
  */
 object KryoSerializer {
 
-  @transient var ser: SparkKryoSerializer = _
+  @transient var ser = new SparkKryoSerializer(SparkEnv.get.conf)
 
   def initWithSharkContext(sc: SharkContext) {
   	ser = new SparkKryoSerializer(sc.sparkEnv.conf)
--- a/src/main/scala/shark/execution/serialization/KryoSerializer.scala
+++ b/src/main/scala/shark/execution/serialization/KryoSerializer.scala
@@ -31,7 +31,7 @@ import shark.SharkContext
  */
 object KryoSerializer {
 
-  @transient var ser = new SparkKryoSerializer(SparkEnv.get.conf)
+  @transient var ser = new SparkKryoSerializer(new SparkConf())
 
   def initWithSharkContext(sc: SharkContext) {
   	ser = new SparkKryoSerializer(sc.sparkEnv.conf)


harveyfeng:
A goal of #234 was to fix an NPE when running SerializationSuite in isolation (i.e. using `sbt/sbt "test-only *Serialization*"`) that your change would re-introduce. One bug with the #234 PR is that it missed a KryoSerializer init that this commit would patch: https://github.com/harveyfeng/shark/commit/dd5a4dcad912a78ffab5dbd817bb8934c6442fb7

Sorry about that whiff :( . What the commit above fix the issue you're running into?

ChengXiangLi:
the case is that, i run shark with spark on yarn, and shark throws error while run any query sql. the NPE error as this pull request description shows is throw on Spark task container. From the shark source code, i find that SharkEnv.init which actually initialize KryoSerializer is only invoked in SharkCliDriver and SharkServer, KryoSerializer is never initialized on Spark task container side, so i assign a initial value for ser in KryoSerializer as shark did before  #234 .
as #234 fix an NPE on SerializationSuite, i think this commit won't re-introduce the issue, as you can see from the diff, i just add an initial value for KryoSerializer.ser, SerializationSuite could still update this value with KryoSerializer.initWithSparkConf.

harveyfeng:
The test suite will still run into an NPE when run isolation, since `SparkEnv.get` will return `null` for the line changed.
(I previously ran `sbt/sbt "test-only *Serialization*"` on your branch, and it failed this way).

The slave-side NPE is definitely an issue, though. What do you think about just initializing a local SparkConf singleton on each slave process? The default SparkConf constructor loads conf vars from system properties, so behavior should be consistent with Shark 0.8.x.

ChengXiangLi:
that make sense, i've update code according to your opinion.

src/main/scala/shark/execution/serialization/KryoSerializer.scala
```diff
@@ -31,7 +31,7 @@ import shark.SharkContext
  */
 object KryoSerializer {
 
-  @transient var ser: SparkKryoSerializer = _
+  @transient var ser = new SparkKryoSerializer(new SparkConf())
```
harveyfeng, line 34: Replace this with

```
@transient lazy val ser: SparkKryoSerializer = {
  val sparkConf = Option(SparkEnv.get).map(_.conf).getOrElse(new SparkConf())
  new SparkKryoSerializer(sparkConf)
}
```

So that SparkConfs don't get initialized twice in the master. We can also remove all the `initWith...` methods (and references in SharkEnv, SerializationSuite, etc).

harveyfeng:
@ChengXiangLi does the code snippet in the comment above look okay?

harveyfeng:
Going to merge this PR and submit another to update a few bits

PR #247, opened 2014-01-22
Delta Compression with better Logging - no boxing

junkumar:


junkumar:
The boxing/unboxing is gone in the delta compression code. The Spire library is used to do this.

junkumar:
#257 brings in these changes.

PR #248, opened 2014-01-23
Added log message to warn user about RDD reload

kayousterhout:
Is this useful to anyone besides me?? If not just ignore this PR, but I found the mystery jobs at the beginning to be confusing w/out an explanatory msg.


rxin:
Jenkins, add to whitelist.

rxin:
Thanks. This is very helpful actually.

PR #249, opened 2014-01-24
Accept spaces before (UN)CACHE, more helpful error message.

kayousterhout:
This is most helpful when running a series of statements
using -e, when adding spaces increases readability.


harveyfeng:
LGTM

PR #253, opened 2014-01-27
Correctly detect in-memory RDDs for table scans.

harveyfeng:


harveyfeng:
note to self: this should be cherry picked into 0.8.1

rxin:
Thanks!

PR #256, opened 2014-02-05
Fixed the bug that all columns for cached tables get pruned in lateral view (and as a result returning nulls for everything).

rxin:


rxin:
I also cherry picked it into master.

PR #261, opened 2014-02-10
Misc updates for Hive 0.11

harveyfeng:
- Detect LATERAL VIEW in ColumnPruner by looking for a LateralViewForwardOperator coupled with a TableScanOperator
- Remove stats\* unit tests from Hive - they're a continuous WIP


harveyfeng:
Jenkins, retest this please!

harveyfeng:
Alright, just need to re-patch windowing functions

src/test/tests_pass.txt
```diff
@@ -766,3 +751,909 @@ testCliDriver_union_lateralview
 
 testCliDriver_union_view
 
+# New from Hive 0.11:
```
rxin: Most of the following tests are not new though?

ChengXiangLi:
looks good, i'm cool with this PR.

src/test/tests_pass.txt
harveyfeng: A *right outer join would be useful here

rxin: note that you can do a sort, and then diff, and then leave only the diff (minus)

harveyfeng: just noticed there's `.q` in some lines, so gotta clean this too

src/main/scala/shark/execution/optimization/ColumnPruner.scala
```diff
@@ -124,8 +124,17 @@ class ColumnPruner(@transient op: TopOperator[_], @transient tbl: Table) extends
     }
 
     // recurse on the subtree
-    op.childOperators.foreach { y =>
-      computeColumnsToKeep(y, cols, op)
+    val numChildren = op.childOperators.size
+    var currentChildIndex = 0
+    while (currentChildIndex < numChildren) {
```
rxin, line 129: Can you add a comment to say what this is doing (i.e. lateral view is a special case blah blah blah).

rxin:
Feel free to merge once you addressed the previous comment.

rxin:
Thanks!

PR #266, opened 2014-02-13
Start the thread local hive session state before running commands to avoid a null pointer in the hive driver code..

marmbrus:


harveyfeng:
@YuryYudin  @apivovarov
Suspect commit: https://github.com/amplab/hive/commit/28e1d8a75371d125813077649af3448420c49cfd

harveyfeng:
HiveServer's clean() gets called after a DDL such as CREATE TABLE. Subsequent commands throw an NPE on the server side because the thread-local SessionState is null because of the cleanse.

Merging this in for now, though we should revert the Hive commit if it's something resolvable in Shark

PR #267, opened 2014-02-13
AstRewrite for count distinct can cause unexpected issues with certain queries

sundeepn:
Fix is to continue without rewrite under any issues in rewriting. Rewrite only when we can.

The exception below is seen for a single alter table add partition command with multiple partitions.

java.lang.ClassCastException: org.antlr.runtime.tree.CommonTree cannot be cast to org.apache.hadoop.hive.ql.lib.Node
at org.apache.hadoop.hive.ql.parse.ASTNode.getChildren(ASTNode.java:61)
at shark.parse.ASTRewriteUtil$.getChildren(ASTRewriteUtil.scala:57)
at shark.parse.ASTRewriteUtil$.findQueryNodes(ASTRewriteUtil.scala:120)
at shark.parse.ASTRewriteUtil$.countDistinctToGroupBy(ASTRewriteUtil.scala:137)
at shark.SharkDriver.compile(SharkDriver.scala:205)
....

### DIFF
--- a/src/main/scala/shark/parse/ASTRewriteUtil.scala
+++ b/src/main/scala/shark/parse/ASTRewriteUtil.scala
@@ -24,6 +24,7 @@ import scala.collection.JavaConversions._
 
 import org.apache.hadoop.hive.ql.lib.Node
 import org.apache.hadoop.hive.ql.parse.{ASTNode, HiveParser}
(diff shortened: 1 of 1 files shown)


harveyfeng:
thanks!

PR #268, opened 2014-02-14
Fix hadoop 0.23.9 compatibility

sundeepn:


project/SharkBuild.scala
```diff
@@ -59,6 +71,10 @@ object SharkBuild extends Build {
   val excludeAsm = ExclusionRule(organization = "asm")
   val excludeSnappy = ExclusionRule(organization = "org.xerial.snappy")
 
+  lazy val protobufVersion = if (isNewHadoop) "2.5.0" else "2.4.1"
+  lazy val akkaVersion = if (isNewHadoop) "2.0.5-protobuf-2.5-java-1.5" else "2.0.5"
+  lazy val akkaGroup = if (isNewHadoop) "org.spark-project" else "com.typesafe.akka"
```
rxin, line 76: I think we can always use the org.spark-project one here.

sundeepn, line 76: I am ok with that. I was just following spark's hadoop 2 configurations.

rxin, line 76: Thanks for clarifying. I didn't realize Spark was doing this also.

PR #279, opened 2014-02-18
get rid of hcatalog dependency

harveyfeng:


harveyfeng:
btw, this needs to be cherry picked into master

PR #280, opened 2014-02-19
Compression fixes

rxin:
1. Fixed a bug that delta encoding was used for every column regardless of compression ratio.
2. Slightly improved performance of compression code. The AMPLab big data benchmark uservisits table used to take 3.5 mins on 5 nodes (20GB/node). With this tiny fix it takes slightly less than 3 mins.


rxin:
This should be applied in the 0.9 release.

junkumar:
Was the loading slowdown due to delta coding usage alone?

rxin:
delta + dictionary

PR #293, opened 2014-02-25
Made _cached MEMORY_ONLY by default.

rxin:


src/main/scala/shark/parse/SharkSemanticAnalyzer.scala
```diff
@@ -465,8 +463,9 @@ class SharkSemanticAnalyzer(conf: HiveConf) extends SemanticAnalyzer(conf) with
       var cacheMode = CacheType.fromString(
         createTableProperties.get(SharkTblProperties.CACHE_FLAG.varname))
       if (checkTableName) {
-        if (tableName.endsWith("_cached")) {
-          cacheMode = CacheType.MEMORY
+        // Use memory only mode for _cached tables, unless the mode is already specified.
+        if (tableName.endsWith("_cached") && cacheMode == CacheType.NONE) {
```
harveyfeng, line 467: For this to mean "shark.cache is unspecified", we might have to add a CacheType.UNSPECIFIED, or do a null check for createTableProperties.get(SharkTblProperties.CACHE_FLAG.varname) ...

Though it doesn't seem like this will come up often, we should still add some docs to clarify semantics

PR #297, opened 2014-02-26
Null mergeCombiners will cause NPE in ExternalAppendOnlyMap

tianyi:
in most time , null mergeCombiners works fine if each key is different
from others, but if some key are duplicate , it will cause NPE when the
function mergeIfKeyExists is called in ExternalAppendOnlyMap.scala


rxin:
Jenkins, test this please.

rxin:
Looks like there's an error in the code that it wouldn't compile. Do you mind updating it?

tianyi:
Done , Sorry for the mistake.

rxin:
Jenkins, retest this please.

rxin:
Jenkins, add to whitelist.

rxin:
Thanks! Merging this now.

PR #300, opened 2014-02-27
AST transformations for DISTINCT aggregates

harveyfeng:
Testing this out. There might be a missing edge case lurking

### DIFF
--- a/src/main/scala/shark/SharkCliDriver.scala
+++ b/src/main/scala/shark/SharkCliDriver.scala
@@ -235,11 +235,19 @@ object SharkCliDriver {
         ret = cli.processLine(line, true)
         prefix = ""
         val sharkMode = SharkConfVars.getVar(conf, SharkConfVars.EXEC_MODE) == "shark"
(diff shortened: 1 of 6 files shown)


lalaguozhe:
I test this feature in Shark branch 0.9, but it give out wrong result. 
My query is "explain select count(distinct queryid) from hive_query_log limit 10". 
When I dont't add limit operator in the query end, the result is correct. 
Through the explain info, the re-writted query creates three stages(MapShuffle->MapShuffle->Result). Both the second stage and third stage contains a limit operator, therefore the second stage  only calculate partial of shuffle data. Is there any fix for this? thx

harveyfeng:
Hey @lalaguozhe, sorry for the huge delay, I'm back on this. We can add 'limit' handling, though my immediate reaction is to just disable rewrites in light of it and maybe warn the user to leave off the 'limit' part. It doesn't make sense to add a 'limit' when 'count distinct' always returns a single number.

lalaguozhe:
Ha, welcome back. I agree with u, limit token should coexist with count(distinct) and throwing an warn to user will be better.

PR #301, opened 2014-03-04
FIX: LazyPTFIterator lose the last row when its partitionKey is different from the previous row's

guowei2:
FIX: LazyPTFIterator lose the last row when its partitionKey is different from the previous row's


rxin:
Jenkins, add to whitelist.

rxin:
Thanks!

PR #302, opened 2014-03-05
Only set the job description instead of also overriding the job group.

marmbrus:
This rolls back a change that I made during the port to 2.10 to avoid using a deprecated Spark API.  However, it looks like that API is going to be un-deprecated and that overriding the job group can break other systems.


rxin:
Thanks!

PR #304, opened 2014-03-07
Merge pull request #43 from sundeepn/fifocache

sundeepn:
Modified FIFO Cache policy - current FIFO cache policy/tests are using gets instead of puts in their cache control logic. So it is behaving as LRU.


harveyfeng:
Thanks for fixing this!

PR #306, opened 2014-03-10
Update run

baishuo:
I think we need classpath when run the script


rxin:
Jenkins, test this please.

rxin:
Jenkins, retest this please.

rxin:
Actually I am pretty sure java picks up the jars from CLASSPATH env variable. Are you sure it was not working for your mysql jars?

baishuo:
Hi @rxin, all jars load by hive? maybe I make a wrong config when the mysql jar didnt work?I am not sure.I dont remeber which version of amplab-hive I used at that time
 I just think we need a classpath when start a java process :)
I also wonder about why the first time test fail?

rxin:
The maven repository was having problems -- that's why the test failed the first time. I don't actually think you need to set classpath here, because they are already set as part of the environmental variable. I'm going to close this one for now. If you find a concrete problem, we can revisit. Thanks!

baishuo:
no problem，thanks @rxin

PR #309, opened 2014-03-13
Update Spark version in build file

kayousterhout:


kayousterhout:
Not sure what the root cause is here but it looks like the github repo for
Spark needs to be updated from incubator

PR #311, opened 2014-03-18
fix "get scala version error"  ---  issue #Shark-219#

CrazyJvm:
${SCALA_HOME}/bin/scala -version | awk '{print $5}' will tell us that the version is "Scala code runner version 2.10.3 -- Copyright 2002-2013, LAMP/EPFL"  rather than "2.10.3"  . Consequently,  the condition "if [ "$SCALA_HOME_VERSION" != "$SCALA_VERSION" ] " will never be true.

issue:https://spark-project.atlassian.net/browse/SHARK-219


CrazyJvm:
Scala 2.10.3 scala -version is outputted to stderr while Scala 2.9.x to stdout. So it performs well in Scala 2.9.x but causes error in Scala 2.10.x.

rxin:
Jenkins, add to whitelist.

rxin:
thanks!

PR #303, opened 2014-03-05
PythonTableRDD

ahirreddy:
Add support for converting TableRDDs to Python compatible RDDs (ByteArrays with pickled python objects as entries)

PR #315, opened 2014-03-28
Add Python SharkContext

brianbrunner:
I've added a working implementation of a python SharkContext. The biggest functionality it exposes is a python version of `sql2rdd`, backed by the PythonTableRDD from https://github.com/amplab/shark/pull/303. 

The only big issue it has currently is that it uses a hack for setting up the Py4j gateway. It injects a bit of code into the `bin/compute-classpath.sh` file in the `SPARK_HOME` directory to add the `HIVE_CONF_DIR` and the assembled shark jar to the Py4j gateway's classpath. The shark jar must be added after the spark jar, so unfortunately we can't just set the `SPARK_CLASSPATH` env variable. It works and does clean up after itself, but overall this method is less than ideal so if someone has a suggestion I'd be willing to rework it.

You should be able to play around with this by running the `./bin/pyshark` script.


mateiz:
Jenkins, test this please

brianbrunner:
I'm closing this due to inactivity. I've added functionality for making an `RDDTable` from Python and I plan on opening a more comprehensive PR at a later date.

PR #316, opened 2014-03-29
Extract Tachyon interface to allow pluggable off-heap storage clients

aarondav:
Right now Tachyon is hard-coded as our off-heap storage format of choice, but users may want to plug in their own faster-than-disk engine. This patch is almost entirely a refactoring of the Tachyon interface into an "OffHeapStorageClient" interface which hides some of the details of the Tachyon implementation from the Shark code.

The interface changes for users of Shark are as follows:
- The OffHeapStorageClient can be specified by setting the "shark.offheap.clientFactory" system property, before creating SharkContext. This is similar to how Spark registers Serializer factories.
- CacheType.TACHYON is deprecated in favor of OFFHEAP.
- We now support `CACHE <TableName> IN <CacheType>` syntax, where CacheType defaults to MEMORY if unspecified.


rxin:
Jenkins, add to whitelist.

src/main/scala/shark/execution/HadoopTableReader.scala
```diff
@@ -17,6 +17,8 @@
 
 package shark.execution
 
+import java.util
```
rxin: probably better to do

```
import java.util.{BitSet => JBitSet}
```

rxin:
Looks like test failures?

src/main/scala/shark/memstore2/OffHeapStorageClient.scala
```diff
@@ -0,0 +1,112 @@
...
+   */
+  lazy val client: OffHeapStorageClient = {
+    try {
+      val clientFactoryClass = Class.forName(clientFactoryClassName)
+      val constructor = clientFactoryClass.getConstructor()
+      val clientFactory = constructor.newInstance().asInstanceOf[OffHeapStorageClientFactory]
+      logInfo("Creating off-heap storage client with factory: " + clientFactoryClassName)
+      clientFactory.createClient()
+    } catch {
+      case e: Exception =>
```
rxin: maybe ClassNotFoundException?

rxin:
also - i think we can remove tachyon_enabled now and just move everything into main. that simplifies the build file a little bit.

src/tachyon_enabled/scala/shark/tachyon/TachyonStorageClient.scala
```diff
@@ -0,0 +1,158 @@
...
+
+import scala.collection.JavaConversions._
+
+import org.apache.spark.rdd.{RDD, UnionRDD}
+import tachyon.client.TachyonFS
+
+import shark.{LogHelper, SharkEnv}
+import shark.execution.serialization.JavaSerializer
+import shark.memstore2.{OffHeapStorageClient, OffHeapStorageClientFactory, TablePartitionStats}
+import java.util
```
rxin: fix the import order

src/test/scala/shark/TachyonSQLSuite.scala
```diff
@@ -44,7 +44,10 @@ class TachyonSQLSuite extends FunSuite with BeforeAndAfterAll {
   var sharkMetastore: MemoryMetadataManager = SharkEnv.memoryMetadataManager
 
   // Determine if Tachyon enabled at runtime.
-  val isTachyonEnabled = SharkEnv.tachyonUtil.tachyonEnabled()
+  val isTachyonEnabled = {
+    val client = OffHeapStorageClient.client
+    client != null && client.getClass.getName == "shark.tachyon.TachyonStorageClient"
```
harveyfeng: This seems to be always true. We could probably get away with checking if the value for TACHYON_MASTER exists in OffHeapStorageClient and avoid instantiating a client in that case.

aarondav:
This patch does not make including Tachyon a requirement, it may still be valid to not want the Tachyon-specific stuff to be in the build?

src/test/scala/shark/TachyonSQLSuite.scala
harveyfeng: (that was also our previous way of checking whether Tachyon is running)

src/main/scala/shark/SharkEnv.scala
```diff
@@ -125,6 +122,7 @@ object SharkEnv extends LogHelper {
   executorEnvVars.put("MESOS_NATIVE_LIBRARY", getEnv("MESOS_NATIVE_LIBRARY"))
   executorEnvVars.put("TACHYON_MASTER", getEnv("TACHYON_MASTER"))
   executorEnvVars.put("TACHYON_WAREHOUSE_PATH", getEnv("TACHYON_WAREHOUSE_PATH"))
+  executorEnvVars.put("SHARK_OFFHEAP_CLIENT_FACTORY", OffHeapStorageClient.clientFactoryClassName)
```
harveyfeng, line 127: Should this be read from the corresponding env variable?

src/main/scala/shark/execution/TableReader.scala
```diff
@@ -63,15 +67,18 @@ class TachyonTableReader(@transient _tableDesc: TableDesc) extends TableReader {
 
   override def makeRDDForTable(
       hiveTable: HiveTable,
-      pruningFnOpt: Option[PruningFunctionType] = None
+      columnsUsed: JBitSet,
+      pruningFn: PruningFunctionType = NonPruningFunction
     ): RDD[_] = {
     val tableKey = MemoryMetadataManager.makeTableKey(_databaseName, _tableName)
-    makeRDD(tableKey, hivePartitionKeyOpt = None, pruningFnOpt)
+    storageClient.readTablePartition(tableKey, None, columnsUsed, pruningFn)
```
harveyfeng: `hivePartitionKeyOpt = None`

src/main/scala/shark/memstore2/OffHeapStorageClient.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  lazy val clientFactoryClassName: String = sys.props.get("shark.offheap.clientFactory")
+    .orElse(sys.env.get("SHARK_OFFHEAP_CLIENT_FACTORY"))
+    .getOrElse("shark.tachyon.TachyonStorageClientFactory")
+
+  /**
+   * Lazily initializes an OffHeapStorageClient, accessible on both the master and worker.
+   * This enables us to avoid serializing the client itself, as we will just create a new one on
+   * each node.
+   * An exception will be thrown if there is no valid shark.offheap.clientFactory specified.
```
harveyfeng: Clarify here whether an exception's thrown every time the client's referenced, or just the first time. Slightly funky use of lazy val ...

src/tachyon_enabled/scala/shark/tachyon/TachyonStorageClient.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  /** We create a new directory with a new RawTable for each independent insert. */
+  private val INSERT_FILE_PREFIX = "insert_"
+
+  /** Non-partitioned tables use a default partition name for consistency. */
+  private val DEFAULT_PARTITION = "_defaultkey"
+
+  private val _fileNameMappings = new ConcurrentJavaHashMap[String, Int]()
+
+  val tfs = if (master != null && master != "") TachyonFS.get(master) else null
```
harveyfeng: Looks like this occurs when Tachyon's built, but the `TACHYON_MASTER` isn't set

src/tachyon_enabled/scala/shark/tachyon/TachyonStorageClient.scala
```diff
@@ -0,0 +1,158 @@
...
+  override def dropTablePartition(tableKey: String, hivePartitionKey: Option[String]): Boolean = {
+    tfs.delete(getPartitionPath(tableKey, hivePartitionKey.getOrElse(DEFAULT_PARTITION)), true)
+  }
+
+  override def readTablePartition(
+      tableKey: String,
+      hivePartitionKey: Option[String],
+      columnsUsed: JBitSet,
+      pruningFn: PruningFunctionType
+      ): RDD[_] = {
```
harveyfeng: spacing

src/tachyon_enabled/scala/shark/tachyon/TachyonStorageClient.scala
```diff
@@ -0,0 +1,158 @@
...
+  override def tablePartitionExists(tableKey: String, hivePartitionKey: Option[String]): Boolean = {
+    tfs.exist(getPartitionPath(tableKey, hivePartitionKey.getOrElse(DEFAULT_PARTITION)))
+  }
+
+  override def dropTable(tableKey: String): Boolean = {
+    tfs.delete(getTablePath(tableKey), true /* recursively */)
+  }
+
+  override def dropTablePartition(tableKey: String, hivePartitionKey: Option[String]): Boolean = {
+    tfs.delete(getPartitionPath(tableKey, hivePartitionKey.getOrElse(DEFAULT_PARTITION)), true)
```
harveyfeng: /\* param name */ for the final `true`

harveyfeng:
It'd be great to establish some invariants on error-handling behavior when the user tries to cache in Tachyon/off-heap storage under these two scenarios:
- Tachyon's built, but not running (i.e., `TACHYON_MASTER` isn't set). The earlier TachyonSQLSuite Jenkins run returned an NPE here.
- Tachyon's not built at all (see comment in OffHeapStorageClient's `lazy val client`)

src/main/scala/shark/memstore2/OffHeapStorageClient.scala
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/** Responsible for creating OffHeapStorageClients. Will be called on master and worker nodes. */
+trait OffHeapStorageClientFactory {
+  def createClient(): OffHeapStorageClient
+}
+
+object OffHeapStorageClient extends LogHelper {
+
+  lazy val clientFactoryClassName: String = sys.props.get("shark.offheap.clientFactory")
```
harveyfeng, line 88: Comment here that this is only used on workers and that the original value is read in SharkEnv on the driver.

src/main/scala/shark/SharkEnv.scala
aarondav, line 127: The name is supposed to come from the System property "shark.offheap.clientFactory" typically, this is just used internally to Shark to transfer the name as an env variable.

src/tachyon_enabled/scala/shark/tachyon/TachyonStorageClient.scala
aarondav: but the line wrapping...

src/main/scala/shark/SharkEnv.scala
harveyfeng, line 127: Reconcile with `shark.offheap.clientFactory` somehow - e.g. set the `SHARK_OFFHEAP_CLIENT_FACTORY` based on that.

src/main/scala/shark/memstore2/OffHeapStorageClient.scala
aarondav, line 88: This is actually used in both places. The SharkEnv thing is to make sure we propagate this to the workers.

src/main/scala/shark/parse/SharkDDLSemanticAnalyzer.scala
```diff
@@ -83,13 +83,25 @@ class SharkDDLSemanticAnalyzer(conf: HiveConf) extends DDLSemanticAnalyzer(conf)
...
-    } else if (newCacheMode == CacheType.MEMORY) {
-      // The table should be cached (and is not already cached).
+
+    if (oldCacheMode == newCacheMode) {
+      logInfo(s"Table is already cached as '$newCacheMode', not changing.")
+      return
+    }
+
+    // Un-cache the table if it's currently cached.
+    // TODO: Could use the cached copy to re-cache the table in a different storage engine.
```
harveyfeng: TODO(aaron)

src/main/scala/shark/memstore2/OffHeapStorageClient.scala
```diff
@@ -0,0 +1,112 @@
...
+
+import shark.LogHelper
+import shark.execution.TableReader.PruningFunctionType
+
+/**
+ * Provides an API for writing to an off-heap storage service, such as Tachyon.
+ *
+ * To ease the distinction between partitioned and non-partitioned tables, we consider a
+ * non-partitioned table as a special case of partitioned tables with the single hivePartitionKey
+ * of value None.
```
harveyfeng: It might be helpful to mention the Tachyon client's `DEFAULT_PARTITION` usage as an example

src/tachyon_enabled/scala/shark/tachyon/TachyonTableRDD.scala
```diff
@@ -45,17 +41,13 @@ private class TachyonTablePartition(rddId: Int, idx: Int, val locations: Seq[Str
 /**
  * An RDD that reads a Tachyon Table.
  */
-class TachyonTableRDD(path: String, @transient sc: SparkContext)
-  extends RDD[TablePartition](sc, Nil) {
+class TachyonTableRDD(path: String, columnsUsed: java.util.BitSet, @transient sc: SparkContext)
```
harveyfeng: java.util.BitSet's already imported - rename to JBitSet?

src/main/scala/shark/execution/TableReader.scala
```diff
@@ -17,44 +17,48 @@
...
 
-/** Helper class for scanning tables stored in Tachyon. */
-class TachyonTableReader(@transient _tableDesc: TableDesc) extends TableReader {
+object TableReader {
+  type PruningFunctionType =
+    (RDD[TablePartition], collection.Map[Int, TablePartitionStats]) => RDD[_]
+}
+
+/** Helper class for scanning tables stored off-heap. */
+class OffHeapTableReader(@transient _tableDesc: TableDesc, storageClient: OffHeapStorageClient)
```
harveyfeng: Any reason why `_tableDesc` has the underscore, but `storageClient` doesn't?

src/main/scala/shark/parse/SharkDDLSemanticAnalyzer.scala
```diff
@@ -83,13 +83,25 @@ class SharkDDLSemanticAnalyzer(conf: HiveConf) extends DDLSemanticAnalyzer(conf)
...
+    oldCacheMode match {
+      case CacheType.MEMORY | CacheType.MEMORY_ONLY =>
+        SharkEnv.memoryMetadataManager.dropTableFromMemory(db, databaseName, tableName)
+      case CacheType.OFFHEAP =>
```
harveyfeng: Nit: { } when case handling takes > 1 line ...

src/tachyon_enabled/scala/shark/tachyon/TachyonOffHeapTableWriter.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import java.nio.ByteBuffer
+
+import tachyon.client.WriteType
+
+import shark.LogHelper
+import shark.execution.serialization.JavaSerializer
+import shark.memstore2.{OffHeapStorageClient, OffHeapTableWriter, TablePartitionStats}
+
+class TachyonOffHeapTableWriter(@transient path: String, @transient numColumns: Int)
```
harveyfeng, line 28: underscore inconsistency here, too

aarondav:
All comments addressed. 

I added proper error handling when TACHYON_MASTER is not set, we now throw an exception upon construction. When Tachyon is not built at all, we currently throw an exception every time we try to use the client. I think this is better behavior than the alternative, which would be to simply throw a NullPointerException each time instead.

src/main/scala/shark/SharkEnv.scala
aarondav, line 127: That's exactly how we're doing it here, is it not?

src/tachyon_enabled/scala/shark/tachyon/TachyonOffHeapTableWriter.scala
harveyfeng, line 28: I guess it's per-file basis, so false alarm :)

aarondav, line 28: Did not address this comment as I'm not sure the overall Shark style guide, but in all Tachyon-related files we aren't using underscores for constructor parameters (so it's not an inconsistency within these files, anyway).

src/main/scala/shark/SharkEnv.scala
harveyfeng, line 127: Ahh okay sorry - it wasn't obvious from the first glance. Add a comment to note it?

harveyfeng:
LGTM

PR #324, opened 2014-04-16
Add a table property to indicate a table being cached

aarondav:
While a table is in the process of being cached in memory or offheap, it is actually inaccessible, because we won't fall back to the base data. This patch puts a band-aid on that by allowing the user to check when the caching is complete, and to disallow multiple cache tasks to execute on the same table at the same time (where the behavior is not defined).

### DIFF
--- a/src/main/scala/shark/parse/SharkDDLSemanticAnalyzer.scala
+++ b/src/main/scala/shark/parse/SharkDDLSemanticAnalyzer.scala
@@ -17,25 +17,20 @@
 
 package shark.parse
 
-import java.util.{HashMap => JavaHashMap}
+import java.util.{HashMap => JavaHashMap, HashSet => JavaHashSet}
 
 import scala.collection.JavaConversions._
 
 import org.apache.hadoop.hive.conf.HiveConf
 import org.apache.hadoop.hive.ql.exec.TaskFactory
-import org.apache.hadoop.hive.ql.parse.ASTNode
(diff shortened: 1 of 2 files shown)


src/main/scala/shark/parse/SharkDDLSemanticAnalyzer.scala
```diff
@@ -115,14 +121,29 @@ class SharkDDLSemanticAnalyzer(conf: HiveConf) extends DDLSemanticAnalyzer(conf)
...
         newCacheMode)
       partSpecsOpt.foreach(partSpecs => sparkLoadWork.partSpecs = partSpecs)
-      rootTasks.head.addDependentTask(TaskFactory.get(sparkLoadWork, conf))
+      val loadTask = TaskFactory.get(sparkLoadWork, conf)
+
+      // Create a task to set CACHE_IN_PROGRESS_FLAG to false once the load completes.
+      val markCachingCompleteDesc = new AlterTableDesc()
+      markCachingCompleteDesc.setOldName(tableName)
+      markCachingCompleteDesc.setOp(AlterTableTypes.ADDPROPS)
+      markCachingCompleteDesc.setProps(new JavaHashMap(Map(
```
harveyfeng, line 139: does this wipe out all other configs?

harveyfeng, line 139: aha it's a putAll

PR #328, opened 2014-05-05
Bug fix: RLDecoder.hasNext may lose duplicated values

liancheng:
When the last several elements of some column are duplicated, `RLDecoder.hasNext` only returns `true` for the first one and loses all those duplicated values.


liancheng:
@rxin Please have a look?

rxin:
Jenkins, add to whitelist.

rxin:
LGTM provided Jenkins is happy.

rxin:
Thanks!

PR #333, opened 2014-05-26
Fix the rdd recovery of table from the non-default database

chenghao-intel:
Previously only support recovery RDD from the default database.

I was trying to change the code like
val cmd = QueryRewriteUtils.cacheToAlterTable("CACHE %s".format(s"$databaseName.$tableName"))

However, seems the Hive doesn't support altering table with database name prefixed.
e.g. "alter table db2.test1 set tblproperties("shark.cache"="memory"); will fail.


rxin:
Thanks. I've merged this.

PR #336, opened 2014-06-01
Fixed a bug that some columns become null for in-memory tables when lateral view explode is used.

rxin:
Column pruning code was too aggressive in pruning out the columns that were actually selected.


aarondav:
LGTM. It appears that Hive hardcodes LateralViewJoins to have a single parent operator which is a Select, which we were not taking into account.

PR #339, opened 2014-06-16
FileAlreadyExistException when writing tachyon table

shimingfei:
currently Shark write data into Tachyon table directly by RAWTable interface in Tachyon.
when  speculation execution is on,   it throws FileAlreadyExistException because two tasks attempts to write same partition into Tachyon table
the approach to solve the problem is first write output of each task into a temp folder,  then commit the result when the task ends. just like write to HDFS.

### DIFF
--- a/project/SharkBuild.scala
+++ b/project/SharkBuild.scala
@@ -56,7 +56,7 @@ object SharkBuild extends Build {
 
   // Whether to build Shark with Tachyon jar.
   val TACHYON_ENABLED = true
-  val TACHYON_VERSION = "0.4.1"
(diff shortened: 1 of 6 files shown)


GraceH:
yes. This fix is quite important in the speculative run. Otherwise, 2 conflict writes are working on the same target. any idea?

haoyuan:
@aarondav can you take a look at this?

haoyuan:
@aarondav , @rxin did this at. :)

aarondav:
LGTM, @rxin please merge!

PR #337, opened 2014-06-04
[WIP] Add CLI Support for Catalyst

chenghao-intel:
- Support reload the cachedRDD upon the start
- Support the CLI switch for Hive/Catalyst

```
$ bin/shark
catalyst> show tables;
Execution Mode: catalyst
OK
shark_test1
shark_test1_cached
Time taken: 0.011 seconds

catalyst> explain select * from shark_test1;
Execution Mode: catalyst
== Logical Plan ==
Project [key#0,val#1]
 MetastoreRelation default, shark_test1, None

== Optimized Logical Plan ==
MetastoreRelation default, shark_test1, None

== Physical Plan ==
HiveTableScan [key#0,val#1], (MetastoreRelation default, shark_test1, None), None
Time taken: 0.172 seconds

catalyst> set shark.exec.mode=hive;
hive> explain select * from shark_test1;
Execution Mode: hive
OK
ABSTRACT SYNTAX TREE:
  (TOK_QUERY (TOK_FROM (TOK_TABREF (TOK_TABNAME shark_test1))) (TOK_INSERT (TOK_DESTINATION (TOK_DIR TOK_TMP_FILE)) (TOK_SELECT (TOK_SELEXPR TOK_ALLCOLREF))))

STAGE DEPENDENCIES:
  Stage-0 is a root stage

STAGE PLANS:
  Stage: Stage-0
    Fetch Operator
      limit: -1
      Processor Tree:
        TableScan
          alias: shark_test1
          Select Operator
            expressions:
                  expr: key
                  type: int
                  expr: val
                  type: string
            outputColumnNames: _col0, _col1
            ListSink


Time taken: 0.107 seconds
```

PR #341, opened 2014-06-20
Spark SQL based Shark CLI

liancheng:
Merged PR #337 authored by @chenghao-intel, and then did some refactoring and update, mainly include:
- Minimized `CatalystContext` as we don't care about response code for now, should update after [SPARK-2106](https://issues.apache.org/jira/browse/SPARK-2106) is addressed.
- `CatalystDriver.destroy` should call `super.destroy`.
- A bunch of trivial coding style fix (including original Shark code).

[Review on Reviewable](https://reviewable.io/reviews/amplab/shark/341)


src/main/scala/org/apache/spark/sql/hive/CatalystContext.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.apache.spark.sql.hive
+
+import org.apache.hadoop.hive.ql.session.SessionState
+import org.apache.spark.SparkContext
+
+import shark.LogHelper
+
+class CatalystContext(sc: SparkContext) extends HiveContext(sc) with LogHelper {
```
rxin, line 25: SparkSQLContext?

src/main/scala/shark/CatalystDriver.scala
```diff
@@ -0,0 +1,89 @@
...
+  override def close(): Int = {
+    hiveResponse = null
+    tableSchema = null
+    0
+  }
+
+  override def getSchema: Schema = tableSchema
+
+  override def getResults(res: JArrayList[String]): Boolean = {
+    if(hiveResponse == null) {
```
rxin: space after if

src/main/scala/shark/CatalystEnv.scala
```diff
@@ -0,0 +1,150 @@
...
+  val activeSessions = new HashSet[String]
+
+  var catalystContext: CatalystContext = _
+  var sparkContext: SparkContext = _
+
+  // The following line turns Kryo serialization debug log on. It is extremely chatty.
+  //com.esotericsoftware.minlog.Log.set(com.esotericsoftware.minlog.Log.LEVEL_DEBUG)
+
+  // Keeps track of added JARs and files so that we don't add them twice in consecutive queries.
+  val addedFiles = HashSet[String]()
```
rxin: are these two still useful?

src/main/scala/shark/SharkCliDriver.scala
```diff
@@ -220,24 +224,32 @@ object SharkCliDriver {
...
+      if (prefix.nonEmpty) {
         prefix += '\n'
       }
       if (line.trim().endsWith(";") && !line.trim().endsWith("\\;")) {
         line = prefix + line
         ret = cli.processLine(line, true)
         prefix = ""
-        val sharkMode = SharkConfVars.getVar(conf, SharkConfVars.EXEC_MODE) == "shark"
-        curPrompt = if (sharkMode) SharkCliDriver.prompt else CliDriver.prompt
+        val sharkMode = SharkConfVars.getVar(conf, SharkConfVars.EXEC_MODE) == "catalyst"
```
rxin: i think we can get rid of the alternative hive exec mode now.

src/main/scala/shark/SharkConfVars.scala
```diff
@@ -0,0 +1,186 @@
...
+
+package shark
+
+import scala.language.existentials
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.hive.conf.HiveConf
+
+
+object SharkConfVars {
```
rxin: i don't think we need this anymore. everything should be in spark sql conf.

README.md
```diff
@@ -59,9 +59,19 @@ resultSet.next()
...
+* Configure the shark_home/conf/shark-env.sh
+* Configure the shark_home/conf/hive-site.xml
+* Start the Shark CLI
+```
+$ bin/shark
+catalyst> show tables;
+catalyst> set shark.exec.mode=hive;
+hive>show tables;
+```
+But there is a bug, which require show tables before doing anything else.
```
chenghao-intel, line 72: This should not be a bug anymore. :)

README.md
```diff
@@ -59,9 +59,19 @@ resultSet.next()
...
+
 ## Known Missing Features
-* Shark CLI
-* Restoring cached tables upon restart
```
chenghao-intel, line 64: This is a missing feature.

src/main/scala/shark/SharkCliDriver.scala
```diff
@@ -246,54 +258,57 @@ object SharkCliDriver {
...
-  if (!ss.isRemoteMode()) {
-    SharkEnv.init()
+  if (!sessionState.isRemoteMode) {
+    CatalystEnv.init()
     if (reloadRdds) {
       console.printInfo(
         "Reloading cached RDDs from previous Shark sessions... (use %s flag to skip reloading)"
         .format(SharkCliDriver.SKIP_RDD_RELOAD_FLAG))
-      TableRecovery.reloadRdds(processCmd(_), Some(console))
+//      TableRecovery.reloadRdds(processCmd(_), Some(console), ss)
```
chenghao-intel: Add TODO here, this is the root where we add back the code of reloading the cached tables.

src/main/scala/shark/SharkConfVars.scala
chenghao-intel: I was thinking if we need to backward-compatible(with to Shark) for the configuration keys name.

rxin: yup we will build it into spark sql to use the right configs. e.g. spark sql should read mapred.reduce.tasks too if the spark sql version of that is not set.

rxin: if you want to give it a try for mapred.reduce.tasks, please go ahead!

src/main/scala/shark/SharkCliDriver.scala
chenghao-intel: That may be helpful if compare the result between Hive & Catalyst, or can we just NOT document it and only for internal usage?

rxin: you can always download hive to use, no?

rxin: either way we can have it, but i am not sure if a single variable justifies a separate config system.

chenghao-intel: 1. Yes, I think the downloaded Hive can be used here as official released Hive, otherwise it should be a bug.
2. Yep, you're right, we'd better get rid of SharkConfVars.scala, as we have the plan to copy all of the configuration key names to spark sql conf.

src/main/scala/org/apache/spark/sql/hive/CatalystContext.scala
marmbrus, line 25: or JDBCContext?

liancheng, line 25: Hmm... I'd rather just call it `SharkContext`. Catalyst is a query optimization framework, Spark SQL is more than that, but neither of them is a concept parallel to Hive. And this class is actually not very much related to JDBC.

marmbrus, line 25: SGTM

README.md
liancheng, line 64: Thanks, the Spark SQL branch is changing somewhat rapidly, and I'd like to update the README file when it's somewhat stabilized :)

liancheng:
@rxin @chenghao-intel, thanks for your comments :) I've removed the alternative Hive execution mode, and addressed issues you brought up. I cleaned up some code that doesn't follow reasonable coding conventions along the way until I realized that most of it was copied from the original Hive codebase. Maybe I should just left them unchanged...

PR #342, opened 2014-06-25
Add support to clear-buffer-cache.py for checking SPARK_HOME for slaves

rnowling:
clear-buffer-cache.py assumes the slaves file is located under /root/spark-ec2/slaves .  This is not consistent with the Spark setup guide for custom deployments.

I added the ability to find the slaves file relative to the SPARK_HOME environmental variable, if set.  If the slaves file cannot be found in either location, an error is raised.

### DIFF
--- a/bin/dev/clear-buffer-cache.py
+++ b/bin/dev/clear-buffer-cache.py
@@ -24,17 +24,18 @@ import sys
 
 EC2_MACHINES_FILE = "/root/spark-ec2/slaves"
 
-spark_home = os.getenv("SPARK_HOME")
-if spark_home is None:
+sparkHome = os.getenv("SPARK_HOME")
+
+machinesFile = None
+if os.path.exists(EC2_MACHINES_FILE):
   machinesFile = EC2_MACHINES_FILE
-  if not os.path.exists(machinesFile):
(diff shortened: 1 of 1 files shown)


bin/dev/clear-buffer-cache.py
```diff
@@ -20,8 +20,22 @@
...
+if spark_home is None:
+  machinesFile = EC2_MACHINES_FILE
+  if not os.path.exists(machinesFile):
+    print "Could not find Spark slaves file.  SPARK_HOME is not set and could not find file at %s" % EC2_MACHINES_FILE
+    sys.exit(1)
+else:
+  machinesFile = spark_home + "/conf/slaves"
+  if not os.path.exists(machinesFile):
+    print "Could not find Spark slaves file. Based on SPARK_HOME, expected file to be located at %s" % machinesFile
+    sys.exit(1)
```
mattf: suggested refactor - lift the test out of the if spark_home is None block, use a generic message

rnowling:
Incorporated mattf's suggestion.  I also updated the variables to use camel case to be consistent with the rest of the script and prioritized the EC2 file location to maintain compatibility with original version of script.

